PR #13, opened 2020-08-14
Fix issue where RPC service hangs when stopped

earlbread:


NineChronicles.Standalone/ActionEvaluationPublisher.cs
```diff
@@ -83,5 +84,11 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
                 stoppingToken
             );
         }
+
+        public override async Task StopAsync(CancellationToken cancellationToken)
+        {
+            await _client.DisposeAsync();
```
longfin, line 90:
```suggestion
            await _client?.DisposeAsync();
```

PR #17, opened 2020-08-14
More detail return status code

riemannulus:
- Return 409: Waiting node service
- Return 503: Unexpected Error
- Return 200: OK


NineChronicles.Standalone/Controllers/GraphQLController.cs
```diff
@@ -44,6 +45,10 @@ public IActionResult RunStandalone()
             try
             {
                 IHostBuilder nineChroniclesNodeHostBuilder = Host.CreateDefaultBuilder();
+                if (StandaloneContext.NineChroniclesNodeService is null)
+                {
+                  return new StatusCodeResult(StatusCodes.Status409Conflict);
```
earlbread, line 50: Shouldn't this be 403 according to the [description](https://github.com/planetarium/NineChronicles.Standalone/pull/17#issue-467876642)?

PR #24, opened 2020-08-27
Restart Swarm<T> if it didn't receive messages

longfin:
This PR causes `LibplanetNodeService.Swarm` to restart if it does not receive a new message within the time specified(`BootstrapInterval`).

### DIFF
--- a/Libplanet.Standalone/Hosting/LibplanetNodeService.cs
+++ b/Libplanet.Standalone/Hosting/LibplanetNodeService.cs
@@ -44,6 +44,8 @@ namespace Libplanet.Standalone.Hosting
 
         private CancellationTokenSource _miningCancellationTokenSource;
 
+        private bool _stopRequested = false;
+
         private static readonly TimeSpan PingSeedTimeout = TimeSpan.FromSeconds(5);
         
         private static readonly TimeSpan FindNeighborsTimeout = TimeSpan.FromSeconds(5);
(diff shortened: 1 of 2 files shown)


Libplanet.Standalone/Hosting/LibplanetNodeService.cs
```diff
@@ -257,6 +198,104 @@ private BaseBlockStatesStore LoadStore(string path, string type, int statesCache
...
+            {
+                Log.Error(e, "Unexpected exception occurred during Swarm.StartAsync(). {e}", e);
+            }
+        }
+
+        private async Task CheckSwarm(CancellationToken cancellationToken = default)
+        {
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                await Task.Delay(BootstrapInterval);
```
dahlia, line 284:
```suggestion
                await Task.Delay(BootstrapInterval, cancellationToken);
```

Libplanet.Standalone/Hosting/LibplanetNodeService.cs
```diff
@@ -257,6 +198,104 @@ private BaseBlockStatesStore LoadStore(string path, string type, int statesCache
...
+            {
+                BootstrapEnded.Set();
+                PreloadEnded.Set();
+            }
+
+            async Task ReconnectToSeedPeers(CancellationToken token)
+            {
+                while (!token.IsCancellationRequested)
+                {
+                await Task.Delay(BootstrapInterval);
```
dahlia, line 252:
```suggestion
                await Task.Delay(BootstrapInterval, token);
```

Libplanet.Standalone/Hosting/LibplanetNodeService.cs
```diff
@@ -120,80 +122,18 @@ public class LibplanetNodeService<T> : IHostedService, IDisposable
...
-                    ReconnectToSeedPeers(cancellationToken)
+                    StartSwarm(preload, cancellationToken),
+                    CheckSwarm(cancellationToken)
                 );
-            }
-            catch (Exception e)
-            {
-                Log.Error(e, "Unexpected exception occurred during Swarm.StartAsync(). {e}", e);
+                preload = false;
+                await Task.Delay(TimeSpan.FromSeconds(15));
```
earlbread, line 133:
```suggestion
                await Task.Delay(TimeSpan.FromSeconds(15), cancellationToken);
```

Libplanet.Standalone/Hosting/LibplanetNodeService.cs
```diff
@@ -120,80 +122,18 @@ public class LibplanetNodeService<T> : IHostedService, IDisposable
...
             }
         }
-
+
```
earlbread, line 136:
```suggestion

```

Libplanet.Standalone/Hosting/LibplanetNodeService.cs
```diff
@@ -120,80 +122,18 @@ public class LibplanetNodeService<T> : IHostedService, IDisposable
...
+            while (!cancellationToken.IsCancellationRequested && !_stopRequested)
             {
                 await await Task.WhenAny(
-                    Swarm.StartAsync(
-                        cancellationToken: cancellationToken,
-                        millisecondsBroadcastTxInterval: 15000
-                    ),
-                    ReconnectToSeedPeers(cancellationToken)
+                    StartSwarm(preload, cancellationToken),
+                    CheckSwarm(cancellationToken)
```
limebell, line 130: What if `CheckSwarm()` throws during IBD?

Libplanet.Standalone/Hosting/LibplanetNodeService.cs
```diff
@@ -257,6 +198,104 @@ private BaseBlockStatesStore LoadStore(string path, string type, int statesCache
...
+                {
+                await Task.Delay(BootstrapInterval);
+                await BootstrapSwarmAsync(0).ContinueWith(t =>
+                {
+                    if (t.IsFaulted)
+                    {
+                    Log.Error(t.Exception, "Periodic bootstrap failed.");
+                    }
+                });
+                token.ThrowIfCancellationRequested();
```
earlbread, line 260:
```suggestion
                    await Task.Delay(BootstrapInterval, token);
                    await BootstrapSwarmAsync(0).ContinueWith(t =>
                    {
                        if (t.IsFaulted)
                        {
                            Log.Error(t.Exception, "Periodic bootstrap failed.");
                        }
                    }, token);
                    token.ThrowIfCancellationRequested();
```

Libplanet.Standalone/Hosting/LibplanetNodeService.cs
earlbread, line 284:
```suggestion
                await Task.Delay(BootstrapInterval, cancellationToken);
```

Libplanet.Standalone/Hosting/LibplanetNodeService.cs
longfin: In IBD, [this check](https://github.com/planetarium/NineChronicles.Standalone/pull/24/files#diff-b2f7c092a968f1c50ebb861d81eaa797R288) will fail(`false`) because `Swarm.LastMessageTimestamp` will be `null`.

but, do you think it would be better if we have an explicit guard for IBD?

limebell: I guess this is enough because `Swarm.LastMessageTimestamp` is only renewed when non-reply messages are received. During IBD, only reply messages are needed.

limebell (approved):

earlbread (approved):

PR #32, opened 2020-09-04
Add HttpRequest to find specific peer on network

limebell:
Added new HttpRequest `/check-peer` that returns 200 if the target peer exists, otherwise 400.

### DIFF
--- a/Libplanet.Standalone/Hosting/LibplanetNodeService.cs
+++ b/Libplanet.Standalone/Hosting/LibplanetNodeService.cs
@@ -165,6 +165,13 @@ namespace Libplanet.Standalone.Hosting
             _miningCancellationTokenSource?.Cancel();
         }
 
+        public async Task<bool> CheckPeer(string addr)
+        {
+            Address address = new Address(addr);
+            var boundPeer = await Swarm.FindSpecificPeerAsync(address, -1);
+            return !(boundPeer is null);
+        }
+
         public Task StopAsync(CancellationToken cancellationToken)
         {
             _stopRequested = true;
--- a/Libplanet.Standalone/Hosting/LibplanetNodeService.cs
+++ b/Libplanet.Standalone/Hosting/LibplanetNodeService.cs
@@ -167,7 +167,7 @@ namespace Libplanet.Standalone.Hosting
 
         public async Task<bool> CheckPeer(string addr)
         {
-            Address address = new Address(addr);
+            var address = new Address(addr);
             var boundPeer = await Swarm.FindSpecificPeerAsync(
                 address, -1, cancellationToken: _swarmCancellationToken);
             return !(boundPeer is null);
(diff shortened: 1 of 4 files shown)


limebell:
pwsh:
```
>> curl -H 'Content-Type: application/json' -X POST http://localhost:12344/check-peer -d @check-peer.json
Found peer D66Cdc360e6d8DF352B9ef99EC65e7BBfC5DC303.
```
`check-peer.json`:
```
{
	"AddressString": "D66Cdc360e6d8DF352B9ef99EC65e7BBfC5DC303"
}
```

Libplanet.Standalone/Hosting/LibplanetNodeService.cs
```diff
@@ -165,6 +165,13 @@ public void StopMining()
             _miningCancellationTokenSource?.Cancel();
         }
 
+        public async Task<bool> CheckPeer(string addr)
+        {
+            Address address = new Address(addr);
+            var boundPeer = await Swarm.FindSpecificPeerAsync(address, -1);
```
earlbread, line 171:
```suggestion
            var boundPeer = await Swarm.FindSpecificPeerAsync(
                address, -1, cancellationToken: _swarmCancellationToken);
```

earlbread (approved):

riemannulus (commented):
It isn't using GraphQL?

limebell:
> It isn't using GraphQL?

No it isn't. Do you think we should use GraphQL?

riemannulus:
IMO Now we implement on GraphQL what something `blockchain specific`. (e.g check tip or check account verify). I think 'check-peer' is also a part of blockchain, like swarm. So maybe we need implement on GraphQL, not http endpoint.

limebell:
> IMO Now we implement on GraphQL what something `blockchain specific`. (e.g check tip or check account verify). I think 'check-peer' is also a part of blockchain, like swarm. So maybe we need implement on GraphQL, not http endpoint.

I think `check-peer` is a feature of network, not really related to blockchain. Is `Swarm`'s specific feature so I thought it should be along with start swarm and set mining features.

riemannulus (approved):
> I think `check-peer` is a feature of network, not really related to blockchain. Is `Swarm`'s specific feature so I thought it should be along with start swarm and set mining features.

Maybe we need align about this 😄 (not this PR)

dahlia (approved):

NineChronicles.Standalone/Controllers/GraphQLController.cs
```diff
@@ -107,6 +110,30 @@ public IActionResult SetMining([FromBody] SetMiningRequest request)
...
+                }
+                
+                return BadRequest($"No such peer {request.AddressString}");
+            }
+            catch (Exception e)
+            {
+                var msg = $"Unexpected error occurred during CheckPeer request. {e}";
+                // Unexpected Error.
+                Log.Warning(e, msg);
+                return BadRequest(msg);
```
moreal, line 133: In my thought, it will match `InternalServerError` than `BadRequest`. Is there `InternalServerError` method in `ControllerBase`? I couldn't find it in its document.

limebell: I've added your suggestion in [`aa4161a`](https://github.com/planetarium/NineChronicles.Standalone/pull/32/commits/aa4161a23f2983a96d27a868f066ea528686dd70)

moreal (approved):

PR #38, opened 2020-09-09
Use LoggedRenderers

limebell:
https://github.com/planetarium/lib9c/pull/51


dahlia (approved):

limebell:
To prevent duplicated logs in block render, I made small change.

PR #58, opened 2020-09-23
Report error via RPC when no any peers in routing table with seed peers given

limebell:
This patch implement interface for reporting errors via RPC, and send `NetworkException` when no any peers in node's routing table even the seed peers were given.

Please see also https://github.com/planetarium/NineChronicles.RPC.Shared/pull/5 and https://github.com/planetarium/NineChronicles.RPC.Shared/pull/6.

### DIFF
--- a/NineChronicles.Standalone/NineChroniclesNodeService.cs
+++ b/NineChronicles.Standalone/NineChroniclesNodeService.cs
@@ -39,6 +39,8 @@ namespace NineChronicles.Standalone
 
         public ActionRenderer ActionRenderer { get; }
 
+        public ExceptionRenderer ExceptionRenderer { get; }
+
         public AsyncManualResetEvent BootstrapEnded => NodeService.BootstrapEnded;
 
         public AsyncManualResetEvent PreloadEnded => NodeService.PreloadEnded;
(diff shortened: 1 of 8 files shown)


longfin:
https://github.com/planetarium/NineChronicles.RPC.Shared/commits/2135be0f3028c13ffa9b7fdb8900a22d89b46ec4 doesn't seem on `main` branch, is it okay?

limebell:
> https://github.com/planetarium/NineChronicles.RPC.Shared/commits/2135be0f3028c13ffa9b7fdb8900a22d89b46ec4 doesn't seem on `main` branch, is it okay?

Fixed! Please review again.

NineChronicles.Standalone/NineChroniclesNodeService.cs
```diff
@@ -148,6 +156,7 @@ public IHostBuilder Configure(IHostBuilder hostBuilder)
                         services.AddHostedService(provider => new ActionEvaluationPublisher(
```
longfin, line 156: We may need to rename this publisher... 😓

longfin (approved):

NineChronicles.Standalone/NineChroniclesNodeService.cs
limebell: Yes I agree. I'll make an issue for this :)

PR #67, opened 2020-09-28
Introduce ActionMutation (CreateAvatar)

riemannulus:


longfin:
How about re-using fixtures and `TableSheetsImporter` in `Lib9c.Tests`?

NineChronicles.Standalone.Tests/TableSheetsImporter.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        public static Dictionary<string, string> ImportSheets()
+        {
+            var sheets = new Dictionary<string, string>();
+            var dir = Path.Combine("Data", "TableCSV");
+            var files = Directory.GetFiles(dir, "*.csv", SearchOption.AllDirectories);
+            foreach (var filePath in files)
+            {
+                var fileName = Path.GetFileName(filePath);
+                if (fileName.EndsWith(".csv"))
```
moreal, line 16: It seems to get files only matched to the pattern, `*.csv` by the line

```csharp
            var files = Directory.GetFiles(dir, "*.csv", SearchOption.AllDirectories);
```

Should it be checked one more time? then I wonder the reason.

NineChronicles.Standalone/GraphTypes/ActionMutation.cs
```diff
@@ -0,0 +1,55 @@
...
+                        var action = new CreateAvatar
+                        {
+                            avatarAddress = avatarAddress,
+                            index = 0,
+                            hair = 0,
+                            lens = 0,
+                            ear = 0,
+                            tail = 0,
+                            name = "createbymutation",
+                        };
```
moreal, line 38: It seems you assumed this mutation must be used only for testing, right?

NineChronicles.Standalone.Tests/NineChronicles.Standalone.Tests.csproj
```diff
@@ -7,6 +7,168 @@
...
+    <None Remove="Data\TableCSV\Skill\BuffSheet.csv" />
+    <None Remove="Data\TableCSV\Skill\EnemySkillSheet.csv" />
+    <None Remove="Data\TableCSV\Skill\SkillBuffSheet.csv" />
+    <None Remove="Data\TableCSV\Skill\SkillSheet.csv" />
+    <None Remove="Data\TableCSV\WeeklyArenaRewardSheet.csv" />
+    <None Remove="Data\TableCSV\WorldAndStage\StageDialogSheet.csv" />
+    <None Remove="Data\TableCSV\WorldAndStage\StageSheet.csv" />
+    <None Remove="Data\TableCSV\WorldAndStage\StageWaveSheet.csv" />
+    <None Remove="Data\TableCSV\WorldAndStage\WorldSheet.csv" />
+    <None Remove="Data\TableCSV\WorldAndStage\WorldUnlockSheet.csv" />
```
moreal, line 49: I wonder what these `None` elements having `Remove` property do though each path seems included as `Content` in the below lines.

NineChronicles.Standalone.Tests/TableSheetsImporter.cs
riemannulus: In fact, this `TableSheetsImporter` is copy from `Lib9c.Tests`. IDK why double-checked file extension. 🤔  I'll test it is ok to remove the `if statement`.

NineChronicles.Standalone/GraphTypes/ActionMutation.cs
riemannulus: Right. In this time, I only assumed using only for testing.

NineChronicles.Standalone.Tests/NineChronicles.Standalone.Tests.csproj
riemannulus: It is auto-created by IDE. 
> By default, the sdk has something like <None Include="**/*"> (simplified) that is added (included) before your project's contents. But you don't want your file to be in the "None" set, but in the "EmbeddedResource" set.

[Reference](https://stackoverflow.com/a/49785443)

longfin (approved):

longfin (approved):

PR #87, opened 2020-10-08
Reference Libplanet inside Lib9c instead of NuGet

dahlia:
See also: <https://github.com/planetarium/lib9c/issues/110> & PR <https://github.com/planetarium/lib9c/pull/126>.


moreal (approved):

moreal (changes requested):
The build seems failed 😢. I guess it is because of `actions/checkout` in the pull_request workflow. Its *submodule* property should be *recursive*.

PR #101, opened 2020-10-20
Add interval between mining in dev mode

limebell:
The line was accidently removed, so restore it.

### DIFF
--- a/NineChronicles.Standalone/NineChroniclesNodeService.cs
+++ b/NineChronicles.Standalone/NineChroniclesNodeService.cs
@@ -159,6 +159,7 @@ namespace NineChronicles.Standalone
                         {
                             Log.Debug("Start mining.");
                             await miner.MineBlockAsync(properties.MaximumTransactions, cancellationToken);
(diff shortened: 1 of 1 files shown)


earlbread (approved):

longfin (approved):

NineChronicles.Standalone/NineChroniclesNodeService.cs
```diff
@@ -159,6 +159,7 @@ public class NineChroniclesNodeService
                         {
                             Log.Debug("Start mining.");
                             await miner.MineBlockAsync(properties.MaximumTransactions, cancellationToken);
+                            await Task.Delay(blockInterval, cancellationToken);
```
moreal, line 162: I wonder it doesn't cost any time for mining in dev mode.

limebell: Its difficulty is at most 2, so it takes almost no time.

moreal (approved):

PR #108, opened 2020-10-22
add push docker image workflow

area363:
I tested the code on my origin repository and then changed the organization name in the yml file to "planetariumhq".


.github/workflows/push_docker_image.yml
```diff
@@ -0,0 +1,21 @@
...
+      - name: checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: login
+        run: docker login --username ${{ secrets.DOCKER_USERNAME }} --password ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+      - name: build
+        run: docker build . -t planetariumhq/ninechronicles-headless:git-${{ github.sha }} --build-arg COMMIT=git-${{ github.sha }}
+      - name: push (publish)
+        run: docker push planetariumhq/ninechronicles-headless:git-${{ github.sha }}
```
limebell, line 21:
```suggestion
        run: docker push planetariumhq/ninechronicles-headless:git-${{ github.sha }}

```

.github/workflows/push_docker_image.yml
```diff
@@ -0,0 +1,21 @@
...
+jobs:
+  build_and_push:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: login
+        run: docker login --username ${{ secrets.DOCKER_USERNAME }} --password ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
```
riemannulus, line 17: Have we added these secrets to this upstream repository?

dahlia: We should add them *before* merging this.

dahlia:
```suggestion
        run: docker login --username '${{ secrets.DOCKER_USERNAME }}' --password '${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}'
```

area363: ok i'll add the secrets right now.

area363:
@planetarium/libplanet Secrets have been added to the upstream repository.

dahlia (approved):

area363:
@planetarium/libplanet I updated the docker workflow to publish git-commit versions as well as the "latest" version so that it would be more convenient for users.

.github/workflows/push_docker_image.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: login
+        run: docker login --username '${{ secrets.DOCKER_USERNAME }}' --password '${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}'
+      - name: build
+        run: docker build . -t planetariumhq/ninechronicles-headless:git-${{ github.sha }} --build-arg COMMIT=git-${{ github.sha }} -t planetariumhq/ninechronicles-headless:latest --build-arg COMMIT=git-${{ github.sha }}
```
moreal, line 19: I had thought since I started to review this PR. The build argument `COMMIT` is used to specify the version suffix of `NineChronicles.Standalone.Executable`. I wonder we need to specify the versioning policy of it.
When I read [https://semver.org](https://semver.org) again, it seems to present build metadata in [another way](https://semver.org/lang/ko/#spec-item-10) but ours haven't. So I thought it will be good that we should discuss it but you can feel that it is early or doesn't need to concern about it. I wonder your opinions @planetarium/libplanet.

moreal, line 19: Though this comment is not related to the above comment, does it need to pass the same build argument twice? 🤔

moreal (approved):

.github/workflows/push_docker_image.yml
area363: @moreal It does appear alittle redundant 😅. We could adjust the Dockerfile so it doesn't require a version suffix as a build argument but I'll bring this issue up tomorrow!

longfin: I guess it would be better if we changed version suffix from `1.0.0-<GIT-SHA>` to `1.0.0+<GIT-SHA>` to follow semver. (even if another PR)

limebell (approved):

longfin (approved):

PR #164, opened 2020-11-13
ValidationQuery returns true when unexpected exception occur

limebell:
SSIA


NineChronicles.Standalone/GraphTypes/ValidationQuery.cs
```diff
@@ -28,12 +28,21 @@ public ValidationQuery(StandaloneContext standaloneContext)
...
                             Thread.Sleep(retryInterval);
                             timeSpent += retryInterval;
+
+                            if (timeSpent < 100 * 1000)
+                            {
+                                continue;
+                            }
+                            
+                            Log.Debug("Blockchain instance is still null.");
+                            return true;
```
longfin, line 43:
```suggestion
                            // FIXME: Should return error instead.
                            return true;
```

longfin (approved):

riemannulus (approved):

dahlia (approved):

PR #166, opened 2020-11-16
Throw when exception occurred during validating metadata

limebell:
Instead of just return true, it became to throw exception instead.

### DIFF
--- a/NineChronicles.Headless/BlockChainInitializeException.cs
+++ b/NineChronicles.Headless/BlockChainInitializeException.cs
@@ -1,0 +1,12 @@
+﻿using System;
+
+namespace NineChronicles.Headless
+{
+    public class BlockChainInitializeException : Exception
+    {
+        public BlockChainInitializeException(string message)
+            : base(message)
+        {
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


riemannulus (approved):

NineChronicles.Standalone/GraphTypes/ValidationQuery.cs
```diff
@@ -25,22 +25,25 @@ public ValidationQuery(StandaloneContext standaloneContext)
...
                             {
                                 continue;
                             }
-                            
-                            Log.Debug("Blockchain instance is still null.");
-                            return true;
+
+                            var msg = $"Blockchain instance is not initialized until {grace}.";
+                            Log.Debug(msg);
+                            throw new NullReferenceException(msg);
```
longfin, line 46: We may need `BlockChainInitializeError` instead of just `NullReferenceException`.

longfin (commented):
I guess it would be better if we throw `ExecutionError` for the GraphQL client. ([see also](https://graphql-dotnet.github.io/docs/getting-started/errors/))

moreal (approved):

longfin (approved):

PR #171, opened 2020-11-16
Filter action by addresses subscribed from client

limebell:
Filters actions using addresses subscribe from the rpc client.

### DIFF
--- a/NineChronicles.Standalone/RPCContext.cs
+++ b/NineChronicles.Standalone/RPCContext.cs
@@ -5,6 +5,6 @@ namespace NineChronicles.Standalone
 {
     public class RpcContext
     {
-        public ImmutableList<Address> AddressesToSubscribe = ImmutableList<Address>.Empty;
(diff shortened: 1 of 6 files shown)


boscohyun (approved):

riemannulus (approved):

Unengine (approved):

ipdae (approved):

NineChronicles.Standalone/RPCContext.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Collections.Immutable;
+using Libplanet;
+
+namespace NineChronicles.Standalone
+{
+    public class RpcContext
+    {
+        public ImmutableList<Address> AddressesToSubscribe = ImmutableList<Address>.Empty;
```
longfin, line 8: I guess it would be better if we use `ImmutableHashSet<>` instead of `ImmutableList<>`.

longfin (approved):

PR #177, opened 2020-11-19
Send Unity error code on NodeException subscribe

riemannulus:
SSIA

### DIFF
--- a/NineChronicles.Standalone/BlockChainService.cs
+++ b/NineChronicles.Standalone/BlockChainService.cs
@@ -10,6 +10,7 @@ using Libplanet.Assets;
 using Libplanet.Blockchain;
 using Libplanet.Blockchain.Renderers;
 using Libplanet.Net;
+using Libplanet.Standalone.Hosting;
 using Libplanet.Tx;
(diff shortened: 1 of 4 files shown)


NineChronicles.Standalone/BlockChainService.cs
```diff
@@ -125,5 +130,18 @@ public UnaryResult<bool> IsTransactionStaged(byte[] txidBytes)
                 isStaged ? "staged" : "not staged");
             return UnaryResult(isStaged);
         }
+
+        public UnaryResult<bool> ReportException(string code, string message)
+        {
+            switch (code)
+            {
+                case "26":
```
moreal, line 138: It should treat the code as shared enum in other pull request.

moreal (approved):

NineChronicles.Standalone/BlockChainService.cs
limebell: I guess we should handle `27` too because it's also action timeout.

limebell (approved):

PR #194, opened 2020-11-25
fix(logging): deal with also GraphQL mode

moreal:
The log stream name will be determined through:

1. Use `standaloneContext.NineChroniclesNodeService.PrivateKey` if it is.
1. Use `standaloneContext.Swarm.Address` if it is.
1. If it doesn't satisfy any above conditions, it will not do anything and sleep 100ms while waiting to be ready to satisfy the conditions.

After, the mechanism will be better to be implemented with `future` having a queue to include log events.

### DIFF
--- a/NineChronicles.Standalone.Executable/AWSSink.cs
+++ b/NineChronicles.Standalone.Executable/AWSSink.cs
@@ -21,7 +21,7 @@ namespace NineChronicles.Standalone.Executable
 
         private AsyncProducerConsumerQueue<LogEvent> _queue;
 
-        public AWSSink(AWSCredentials credentials, RegionEndpoint regionEndPoint, Func<string> logGroupNameGetter, Func<string> logStreamNameGetter = null)
+        public AWSSink(AWSCredentials credentials, RegionEndpoint regionEndPoint, string logGroupName, string logStreamName)
         {
             Config = new AmazonCloudWatchLogsConfig();
             Config.RegionEndpoint = regionEndPoint;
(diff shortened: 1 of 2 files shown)


NineChronicles.Standalone.Executable/AWSSink.cs
```diff
@@ -56,29 +56,39 @@ private async Task Worker(CancellationToken cancellationToken)
...
-                string sequenceToken = await GetSequenceToken(logGroupName, logStreamName);
-
-                cancellationToken.ThrowIfCancellationRequested();
-                LogEvent logEvent = await _queue.DequeueAsync(cancellationToken);
-                var request = new PutLogEventsRequest(logGroupName, logStreamName, new List<InputLogEvent>
-                {
-                    new InputLogEvent
+                    string logGroupName = LogGroupNameGetter();
+                    string logStreamName = LogStreamNameGetter();
+                    Console.Error.WriteLine($"logGroupName: {logGroupName} / logStreamName: {logStreamName}.");
```
dahlia, line 63: This seems a debug print.  Is it intended to be in the commit?

moreal: Actually, I intended it for later debugging but it is okay to remove if others feel useless.

riemannulus: IMO It is ok for debugging later.

moreal:
While I had this job, I became to wonder how can we distinguish the context of a user. The private key (address) will be generated randomly and it will not have any relations with the user. Then, what will we ask something to distinguish the context, and how can we get information from the logs? So I want to suggest random GUID as the standalone's logging session-id. or I hope to leave another idea if there is someone having a better idea? @planetarium/libplanet

dahlia:
Giving users a persistent UUID would be good.

moreal:
As a summary of changes:

1. It doesn't separate logs by date for two goals.
    1. Do not require permission to create a log group.
    1. To see logs in one sequential view. Before this, if you want to see logs in *2020.11.25 23:50 ~ 2020.11.26 00:10*, you must see logs after you open another view of the next day because they are not connected smoothly.
1. Use UUID stored at the specific path as a log stream name to search logs by persistent criteria.

limebell (approved):
👍

longfin (approved):

PR #197, opened 2020-11-27
feat(logging): support Cognito identity credential

moreal:
It supports [AWS Cognito credentials](https://aws.amazon.com/ko/cognito/).

To use this feature, you *MUST* use only one crendtial method between basic credential (i.e., access key & secret key) and Cognito credential (i.e., Cognito identity).


longfin (approved):

limebell:
Maybe we need error logging for the case where both basic credential and Cognito credential command line options were given.

moreal:
> 
> 
> Maybe we need error logging for the case where both basic credential and Cognito credential command line options were given.

Thanks for your opinion, @limebell. I believe your opinion was applied at 63652e8, please check it 😀

dahlia (approved):

moreal:
@limebell Could you review again, please? 🙏

limebell (approved):

PR #216, opened 2020-12-15
feat(gql): add avatar state query

moreal:
## Overview

It provides [`Lib9c.Model.State.AvatarState`][lib9c-avatar-state] as GraphQL API.

## Comment about a test

I added a test for only `CollectionMapType` because I thought other types are covered from tests of graphql-dotnet/graphql-dotnet. For instance, `StatTypeEnumType` is actually the alias of `EnumerationGraphType<StatType>` and if I wrote its test code, it will just check that `new StatTypeEnumType().Serialize(StatType.*)` evaluate as the capital of the enum. I thought there is no duty to test them because they should be tested in `GraphQL.Tests` and guaranteed. If it seems that I am thinking in an incorrect way or you feel like to want discussion for this, please leave a comment.

[lib9c-avatar-state]: https://github.com/planetarium/lib9c/blob/main/Lib9c/Model/State/AvatarState.cs

### DIFF
--- a/NineChronicles.Headless.Tests/GraphQLTestUtils.cs
+++ b/NineChronicles.Headless.Tests/GraphQLTestUtils.cs
@@ -1,0 +1,29 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using GraphQL;
+using GraphQL.Types;
+
+namespace NineChronicles.Headless.Tests
+{
+    public static class GraphQLTestUtils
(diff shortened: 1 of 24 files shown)


riemannulus (approved):

moreal:
@planetarium/libplanet PTAL. 🙏

dahlia (approved):

PR #234, opened 2021-01-08
feat: provide authorized miner mode

moreal:
It provides `--is-authorized-miner` flag for *authorized miner* mode. The mode makes miner mine only blocks that authorized miner should mine.

PR #240, opened 2021-01-12
fix: correct index to mine by authorized miner

moreal:
It fixes a bug that it didn't detect the index of the block to mine correctly. It was appeared since #234.


moreal:
@planetarium/libplanet PTAL 🙏

limebell (approved):

dahlia (approved):

PR #244, opened 2021-01-14
fix(logging): cover situation that guid is broken

moreal:
Currently, it loads the `Guid` used in `AWSSink` and store it every run. Maybe it may make the `Guid` if `NineChronicles.Headless` killed after short time. Also it hasn't covered the situtation. I believe This pull request resolves them.

### DIFF
--- a/NineChronicles.Headless.Executable/Program.cs
+++ b/NineChronicles.Headless.Executable/Program.cs
@@ -352,7 +352,7 @@ namespace NineChronicles.Headless.Executable
             string path = AWSSinkGuidPath();
             if (!File.Exists(path))
             {
-                Console.Error.WriteLine($"AWSSink id doesn't exist.");
+                Console.Error.WriteLine($"AWSSink id doesn't exist. (path: {path})");
                 return null;
             }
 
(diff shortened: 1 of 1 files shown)


NineChronicles.Headless.Executable/Program.cs
```diff
@@ -346,7 +350,20 @@ static void ConfigureSentryOptions(SentryOptions o)
         private Guid? LoadAWSSinkGuid()
         {
             string path = AWSSinkGuidPath();
-            return File.Exists(path) ? Guid.Parse(File.ReadAllText(AWSSinkGuidPath())) : (Guid?)null;
+            if (!File.Exists(path))
+            {
+                Console.Error.WriteLine($"AWSSink id doesn't exist.");
```
longfin, line 355: I guess displaying `path` is useful.

```suggestion
                Console.Error.WriteLine($"AWSSink id doesn't exist. (path: {path})");
```

longfin (approved):

longfin (approved):

limebell (approved):

PR #249, opened 2021-01-22
Fix NullReferenceException with empty authorized miner state

limebell:
Prevent `NullReferenceException` with empty authorized miner state in block policy.


longfin (approved):

moreal (approved):
As a little opinion, I'm not sure it is okay the situation that, `--authorized-miner` was set but there is no `AuthorizedMinerState`. I think it will be better to notice it to the user and stop the application.

area363 (approved):

riemannulus (approved):

PR #273, opened 2021-02-02
feat: send cloudwatch logs periodically

moreal:
Currently, there is a problem with memory leaks because of `AWSSink._queue`. There was a time which couldn't handle `LogEvent`s well so the capacity of the `Queue<T>` used by _queue, became more and more, up to 4GB in my case.

This pull request tries to resolve the problem by handling the events at the correct time well with `PeriodicBatchSink`. You can see the sink's more information in [serilog/serilog-sinks-periodicbatching](https://github.com/serilog/serilog-sinks-periodicbatching). I opened this because the test in local was better than now when monitoring with memory profiler. I believe it will make the memory leak problem better but I'm not sure that it can fix all memory leaks well 🤔 .

If you want to test this in your local, make sure you must run with `--aws-*` options.

### DIFF
--- a/NineChronicles.Headless.Executable/AWSSink.cs
+++ b/NineChronicles.Headless.Executable/AWSSink.cs
@@ -1,36 +1,31 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Amazon;
 using Amazon.CloudWatchLogs;
 using Amazon.CloudWatchLogs.Model;
 using Amazon.Runtime;
-using Nito.AsyncEx;
-using Serilog.Core;
 using Serilog.Events;
+using Serilog.Sinks.PeriodicBatching;
 
 namespace NineChronicles.Headless.Executable
 {
-    public class AWSSink : ILogEventSink, IDisposable
+    public class AWSSink : IBatchedLogEventSink, IDisposable
     {
         private readonly AmazonCloudWatchLogsClient _client;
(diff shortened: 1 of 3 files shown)


ipdae (approved):

moreal:
PTAL @longfin @area363

area363 (commented):
@moreal Thanks, I was checking it out on my local.
image
Are these Amazon CloudwatchLogs exceptions acceptable when running the headless application? These seem to appear most towards the beginning of the program execution.

moreal:
> 
> 
> @moreal Thanks, I was checking it out on my local.
> image
> Are these Amazon CloudwatchLogs exceptions acceptable when running the headless application? These seem to appear most towards the beginning of the program execution.

How did you run the headless? On command line? Or Rider debug?

Could you let me know your running environment (e.g. OS)

moreal:
> How did you run the headless? On command line? Or Rider debug?
> 
> Could you let me know your running environment (e.g. OS)

I received it internally.

moreal:
@area363 I tried to find the real reason to cause the messages but I couldn't find the correct one 😞 But, in my thought, it seems to logs for debugging.

area363:
> @area363 I tried to find the real reason to cause the messages but I couldn't find the correct one 😞 But, in my thought, it seems to logs for debugging.

@moreal Right, I also think it's just a debug log. The program itself worked as expected. Thanks for the clarification!

area363 (approved):

area363 (approved):

PR #304, opened 2021-02-17
add height to statequery

area363:
This PR takes a `block hash` as an offset for [GetState()](https://github.com/planetarium/libplanet/blob/95a8409bf11017d742be14ff19e49b913ceaeb42/Libplanet/Blockchain/BlockChain.cs#L490).
Screen Shot 2021-02-17 at 7 15 02 PM
The code seems a little verbose(especially the `if else` logic) so I'm open to suggestions 😄

### DIFF
--- a/NineChronicles.Headless/GraphTypes/StateQuery.cs
+++ b/NineChronicles.Headless/GraphTypes/StateQuery.cs
@@ -1,17 +1,15 @@
-using System.Security.Cryptography;
 using Bencodex.Types;
 using GraphQL;
 using GraphQL.Types;
 using Libplanet;
 using Libplanet.Action;
-using Libplanet.Blockchain;
 using Nekoyume;
 using Nekoyume.Model.State;
 using NineChronicles.Headless.GraphTypes.States;
 
 namespace NineChronicles.Headless.GraphTypes
 {
-    public class StateQuery<T> : ObjectGraphType<BlockChain<T>>
+    public class StateQuery<T> : ObjectGraphType<AccountStateGetter>
         where T : IAction, new()
     {
         public StateQuery()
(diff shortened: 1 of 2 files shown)


NineChronicles.Headless/GraphTypes/StateQuery.cs
```diff
@@ -18,42 +19,106 @@ public StateQuery()
...
+                    new QueryArgument<ByteStringType>
+                    {
+                        Name = "hash",
+                        Description = "Offset block hash for query."
+                    }),
                 resolve: context =>
                 {
                     var address = context.GetArgument<Address>("address");
-                    return new AvatarState((Dictionary)context.Source.GetState(address));
+                    byte[] blockHashByteArray = context.GetArgument<byte[]>("hash");
```
longfin, line 35: Perhaps, we may need to add a primitive type for `HashDigest<SHA256>` like `Address` to avoid duplications 🤔

longfin (approved):

moreal (commented):
As you left at the pull request description, It seems to have many code duplications to get state by `hash`. I suggest to resolve this problem with `AccountStateGetter`. `StateQuery` doesn't need all functions of `BlockChain<T>` and *only get state*. Then, in my thought, the `ObjectGraphType<BlockChain<T>>` can be `ObjectGraphType<AccountStateGetter>` without generic type parameter in natural.

`AccountStateGetter`'s definition is same with the below line:

```csharp
    public delegate IValue? AccountStateGetter(Address address);
```

There is no argument to pass the block `hash`. So it should be passed in `StandaloneQuery` level. So I tried with below lines..:

```csharp
Field<NonNullGraphType<StateQuery<NCAction>>>(name: "stateQuery", arguments: new QueryArguments(
    new QueryArgument<ByteStringType>
    {
        Name = "hash",
    }),
    resolve: context =>
    {
        HashDigest<SHA256>? blockHash = context.GetArgument<byte[]>("hash") switch
        {
            byte[] bytes => new HashDigest<SHA256>(bytes),
            null => null,
        };

        IValue? GetState(Address address) =>
            standaloneContext.BlockChain.GetState(
                address,
                blockHash);

        return (AccountStateGetter)GetState;
    });
```

I hope able to query like below an example query and I believe it will be better then 😉 

```graphql
query {
  stateQuery(hash: "<HASH>") {
    avatar { ... }
    shop { ... }
  }
}
```

area363:
Below is the modified query. This way, you only need to insert a block hash argument once.

Screen Shot 2021-02-18 at 5 05 24 PM

NineChronicles.Headless/GraphTypes/StateQuery.cs
```diff
@@ -3,14 +3,13 @@
...
 using Libplanet.Action;
-using Libplanet.Blockchain;
 using Nekoyume;
 using Nekoyume.Model.State;
 using NineChronicles.Headless.GraphTypes.States;
 
 namespace NineChronicles.Headless.GraphTypes
 {
-    public class StateQuery<T> : ObjectGraphType<BlockChain<T>>
+    public class StateQuery<T> : ObjectGraphType<AccountStateGetter>
```
moreal, line 12: It seems not needed `T` type parameter to me. Are there lines to use it?

area363: You're right, thanks for catching that! I'll fix that part.

moreal (approved):

longfin (approved):

ipdae (approved):

PR #309, opened 2021-02-19
Avoid NRE in minerLoopAction

ipdae:
SSIA


moreal (approved):
It's not related to this pull request changes but it seems we should use `chain.Policy` instead of `blockPolicy`. Thanks to fix the bug! 🙏

area363 (approved):

limebell:
Seems like duplication of #249

PR #367, opened 2021-03-16
feat(cli): introduce conversion subcommand

moreal:
This pull request introduces `conversion private-key` subcommand to replace `query.keyStore.privateKey` GraphQL query.


NineChronicles.Headless.Executable/Program.cs
```diff
@@ -29,6 +29,7 @@ namespace NineChronicles.Headless.Executable
     [HasSubCommands(typeof(ValidationCommand), "validation")]
     [HasSubCommands(typeof(ChainCommand), "chain")]
     [HasSubCommands(typeof(KeyCommand), "key")]
+    [HasSubCommands(typeof(ConversionCommand), "conversion")]
```
longfin, line 32: I guess `conversion` features relate to `key`, so it's more natural for me that integrate with `key`. (or merging them into `account`)

moreal: How way do you think to merge them? Can you give me an example? In my thought, there is no relationship because it does only conversion without dependency with key store.

longfin: > there is no relationship because it does only conversion without dependency with key store.

Technically yes, they don't have any relationship. but from the user-side, they want to convert private key to address (or reverse) when managing their Keystore IMO. considering this context, `key` and `conversion` features can be related.

> Can you give me an example?

I imagined like as following...

```bash
$ NineChronicles.Headless.Executable key convert --address <private-key> 
$ NineChronicles.Headless.Executable key convert --public-key ​<private-key>
```

Also, even though features belonging to `conversion` is not directly related to `key`, I think the name `conversion` is too general.

moreal: Thanks to describe it. And I understood. Last one, I wondered one point, do you belive this command structure can be applied also to `Libplanet.Tools`?

longfin: > do you belive this command structure can be applied also to Libplanet.Tools?

Yes 😅

moreal: I applied it in 1e1bfe8 and 5679a22. Could you check it, please? 🙏

moreal: Sadly, the logic was wrong, I will comment you after I fix.

ipdae (approved):

NineChronicles.Headless.Executable/Program.cs
moreal: Could you review again, please? You can see 1e1bfe8, 5afdd48.

README.md
```diff
@@ -24,6 +24,7 @@ Commands:
   validation
   chain
   key
+  conversion
```
longfin, line 27: is it needed?

moreal: No, it doesn't need after `conversion` merged into under `key` subcommand. I will remove it.

moreal: I removed it in 6acc9ac.

longfin (approved):

PR #370, opened 2021-03-16
Prevent swarm restarting

limebell:
In some situation swarm is started repeatedly where it is already running. This patch prevents it by removing loop in `Service.StartAsync()`.


longfin (approved):

longfin:
It needs to be rebased.

PR #381, opened 2021-03-18
refactor: enable nullable feature over all projects

### LINKED ISSUE #349: Enable nullable feature over projects
Since C# 8.0, *Nullable references* has been introduced and made able to control `null` more efficiently. So I suggest to enable the feature over all projects in this repository, by adding this line into each *.csproj* files.

```XML
<Nullable>enable</Nullable>
```

You can see more https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references.

moreal:
It resolves #349.


NineChronicles.Headless/GraphTypes/StandaloneSubscription.cs
```diff
@@ -39,9 +39,16 @@ class PreloadStateType : ObjectGraphType<PreloadState>
...
             private class PreloadStateExtra
             {
-                public string Type { get; set; }
-                public long CurrentCount { get; set; }
-                public long TotalCount { get; set; }
+                public string Type { get; }
+                public long CurrentCount { get; }
+                public long TotalCount { get; }
+
+                public PreloadStateExtra(string type, long currentCount, long totalCount)
```
longfin, line 46: Why we need this constructor instead of previous property setters?

NineChronicles.Headless/GraphTypes/Notification.cs
```diff
@@ -5,13 +5,9 @@ public class Notification
         public NotificationEnum Type { get; set; }
         public string Message { get; set; }
 
-        public Notification(NotificationEnum type)
-        {
-            Type = type;
-        }
-
-        public Notification(NotificationEnum type, string msg) : this(type)
+        public Notification(NotificationEnum type, string msg = "")
```
longfin, line 8: Can we use `string.Empty`?

NineChronicles.Headless/GraphTypes/StandaloneSubscription.cs
moreal: The properties should be `non-nullable`, but, if we uses the property setter you said, it can't guarantee that it will be set always then it should be `nullable`. But the situtation seems not intended.

image

NineChronicles.Headless/GraphTypes/Notification.cs
moreal: No, we can't. Because `string.Empty` is not contant value, it can't be used as `default` value.

ipdae (approved):

area363 (approved):
Branch needs to be updated!

PR #413, opened 2021-04-02
feat(graphql): replace miner operation HTTP endpoints

moreal:
# AS-IS

1. Call `/set-private-key` endpoint to set miner's private key.
1. Call `/set-mining` endpoint to start mining.

# TO-BE

1. Run headless with `--graphql-admin-passphrase` option, the secret passphrase to prove the user is admin.
1. Call `auth.login` mutation with `adminPassphrase` argument to register your private key to session and to sign in as admin.
1. Call `miner.start` mutation.


ipdae (approved):

area363 (approved):

longfin:
It seems to have conflicts.

PR #464, opened 2021-05-03
add error log to `LibplanetNodeService<T>`

area363:
This PR logs the overall error messages when `LibplanetNodeService` is disposed during preloading.

### DIFF
--- a/Libplanet.Headless/Hosting/LibplanetNodeService.cs
+++ b/Libplanet.Headless/Hosting/LibplanetNodeService.cs
@@ -348,26 +348,29 @@ namespace Libplanet.Headless.Hosting
                     _preloadStatusHandlerAction(true);
                     try
                     {
-                        await Swarm.PreloadAsync(
-                            TimeSpan.FromSeconds(5),
-                            PreloadProgress,
-                            cancellationToken: cancellationToken
-                        );
-                    }
-                    catch (AggregateException e)
-                    {
-                        if (!_ignorePreloadFailure)
+                        try
                         {
-                            throw;
+                            await Swarm.PreloadAsync(
+                                TimeSpan.FromSeconds(5),
+                                PreloadProgress,
+                                cancellationToken: cancellationToken
+                            );
                         }
+                        catch (AggregateException e)
+                        {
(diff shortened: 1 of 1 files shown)


Libplanet.Headless/Hosting/LibplanetNodeService.cs
```diff
@@ -363,6 +363,14 @@ Task BootstrapSwarmAsync(int depth)
 
                         Log.Error(e, "{Message}", e.Message);
                     }
+                    catch (Exception e)
+                    {
+                        Log.Error(
+                            e,
+                            "An exception occurred during preload async: {Message}",
```
moreal, line 370: I wonder the message should have `async` description. If you wanted to mention `Swarm.PreloadAsync`, in my opinion, it seems better to use `nameof` keyword. Like...

```suggestion
                            $"An exception occurred during {nameof(Swarm.PreloadAsync)}: {{Message}}",
```

Or without `async` mention...

```suggestion
                            "An exception occurred during preloading: {Message}",
```

-----

Also it seems unexpected exception. So, how about adding the *unexpected* adjective?

```suggestion
                            "An unexpected exception occurred during preload async: {Message}",
```

Libplanet.Headless/Hosting/LibplanetNodeService.cs
```diff
@@ -363,6 +363,14 @@ Task BootstrapSwarmAsync(int depth)
 
                         Log.Error(e, "{Message}", e.Message);
                     }
+                    catch (Exception e)
```
longfin, line 366: It can't catch from [rethrowing](https://github.com/planetarium/NineChronicles.Headless/pull/464/files#diff-99583c6ac600be724a6dcad7bad852f486e53190003ccaa7f17afc153b80b81bL361). we need to nest try-catch block if want to catch exceptions from there.

moreal (approved):

ipdae (approved):

area363:
Closing this PR because it's already merged in [v100043 commit](https://github.com/planetarium/NineChronicles.Headless/commit/e43d3b6ba49d3160c29b005298789051657898e0).

PR #469, opened 2021-05-11
fix(graphql): use `BlockRenderer`'s subject directly

moreal:
This pull request fixes `subscription.tipChanged` doesn't work and expects a small performance improvement by using `BlockRenderer.BlockSubject` directly, skipping one step.


moreal:
#481 pull request is trying to fix the duplicated issue, so I will close this pull request for avoiding useless conflicts.

PR #487, opened 2021-05-24
Feature/calculate reward info

ipdae:


NineChronicles.Headless/GraphTypes/StandaloneSubscription.cs
```diff
@@ -218,9 +221,14 @@ private void RenderBlock((Block<PolymorphicAction<ActionBase>> OldTip, Block<Pol
                 {
                     MonsterCollectionState monsterCollectionState = new MonsterCollectionState(collectDict);
                     canReceive = monsterCollectionState.CanReceive(pair.NewTip.Index);
+                    var rewardLevel= monsterCollectionState.GetRewardLevel(pair.NewTip.Index);
+                    if (rewardLevel > 0)
+                    {
+                        rewardInfos = monsterCollectionState.RewardLevelMap[rewardLevel];
+                    }
```
riemannulus, line 228: Maybe we can use the same code snippet on the query

```suggestion
                            var rewardInfos = rewardLevel > 0
                                ? monsterCollectionState.RewardLevelMap[rewardLevel]
                                : new List<MonsterCollectionRewardSheet.RewardInfo>();
```

longfin (approved):

NineChronicles.Headless/GraphTypes/StandaloneSubscription.cs
ipdae: If `Agent` or `MonsterCollectionState` is null, NRE can occur, so the default value is defined in line 215.

PR #491, opened 2021-05-26
ReducedStore

dahlia:
In order to workaround the issue addressed in <https://github.com/planetarium/libplanet/issues/1313>, I added `ReducedStore` which omits `TxSuccess.UpdatedStates` at all.

### DIFF
--- a/Libplanet.Headless/ReducedStore.cs
+++ b/Libplanet.Headless/ReducedStore.cs
@@ -1,0 +1,153 @@
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using Bencodex.Types;
+using Libplanet.Action;
+using Libplanet.Blocks;
+using Libplanet.Store;
+using Libplanet.Tx;
+
+namespace Libplanet.Headless
+{
+    /// <summary>
+    /// A <see cref="IStore"> decorator that reduce space consumption by omitting input calls which
+    /// are unused by Nine Chronicles.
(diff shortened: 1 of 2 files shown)


Libplanet.Headless/ReducedStore.cs
```diff
@@ -0,0 +1,153 @@
...
+            InternalStore.GetBlockPerceivedTime(blockHash);
+
+        public Guid? GetCanonicalChainId() =>
+            InternalStore.GetCanonicalChainId();
+
+        public Transaction<T> GetTransaction<T>(TxId txid) where T : IAction, new() =>
+            InternalStore.GetTransaction<T>(txid);
+
+        public TxExecution GetTxExecution(BlockHash blockHash, TxId txid) =>
+            InternalStore.GetTxExecution(blockHash, txid);
```
longfin, line 86: How about throwing `NotSupportedException`?

longfin (approved):
LGTM

Libplanet.Headless/ReducedStore.cs
dahlia: We still need `TxFailure`, even though `TxSuccess` is unused.

longfin: 🆗

dahlia:
@area363 Could we give it a try to see if it actually is effective in reducing space consumption?

PR #522, opened 2021-06-21
Use static peer in Liplanet

limebell:
This patch use static peer implemented in Libplanet, and remove previous static peer checking task. Use `--static-peer {peer}` to use this option.


limebell:
PTAL

riemannulus (approved):

longfin (approved):

PR #526, opened 2021-06-22
Fix to broadcast right away

longfin:
This PR forces `stageTx()` mutation to broadcast staged tx right away.


ipdae (approved):

area363 (approved):

longfin:
Fixed a bug where GQL had returned exception when `service.swarm` hadn't started. PTAL @area363 @ipdae

ipdae (approved):

area363 (approved):

PR #536, opened 2021-06-25
renewal shop query

ipdae:
- update shop API by https://github.com/planetarium/lib9c/pull/465
- open after https://github.com/planetarium/lib9c/pull/465


ipdae:
renewal shopState query & update sell, buy mutation

PR #539, opened 2021-06-29
Release v100052

longfin:


area363:
Do we usually do /rebase in this case?

longfin:
> Do we usually do /rebase in this case?

`rc` → `development` means we want to apply the same hotfix as `main` to the dev branch. in that case, I prefer to merge over rebase because it doesn't hurt any other commits already in `development`.

PR #323, opened 2021-02-24
feat(graphql): introduce libplanet-explorer as `chainQuery'

moreal:
This pull request introduces `chainQuery` which equals with [libplanet-explorer](https://github.com/planetarium/libplanet-explorer), as a step of #298. This query will provide queries like below screenshot as libplanet-explorer does. And it will replace the other queries (e.g. `getTx`, `nodeStatus.tip`, etc) in next pull requests step by step.

image

## What this pull request changes

- Introduce new query `chainQuery`, libplanet-explorer's root `Query` as a sub-query.
- Remove `GraphType`s copied from libplanet-explorer or duplicated, to replace them with the `GraphType`s in libplanet-explorer.
  - `AddressType`
  - `ByteStringType`
  - `ActionType`

PR #542, opened 2021-06-30
add offset parameter to topmostblock query

area363:
This PR adds an `offset` parameter to the `topmostblock` query so that the launcher can utilize it to show confirmed blocks mined by players instead of showing the blocks mined from the chain tip which are prone to `reorg`.

### DIFF
--- a/NineChronicles.Headless.Tests/GraphTypes/StandaloneQueryTest.cs
+++ b/NineChronicles.Headless.Tests/GraphTypes/StandaloneQueryTest.cs
@@ -217,6 +217,62 @@ namespace NineChronicles.Headless.Tests.GraphTypes
             Assert.Equal(expectedResult, result.Data);
         }
 
+        [Fact]
+        public async Task NodeStatusGetTopMostBlocks()
+        {
+            var userPrivateKey = new PrivateKey();
+            var userAddress = userPrivateKey.ToAddress();
+            var service = MakeMineChroniclesNodeService(userPrivateKey);
+            StandaloneContextFx.NineChroniclesNodeService = service;
+            StandaloneContextFx.BlockChain = service.Swarm?.BlockChain;
+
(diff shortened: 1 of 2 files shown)


ipdae (approved):

moreal (commented):
I have question for this feature 🤔

Since #323, the `chainQuery.blockQuery.blocks` query was introduced and it may have parameters like `desc`, `offset` and `limit`. IMO it seems the query of the libplanet-explorer can replace `topmostBlocks` query. 🤔 How do you think?

area363:
> I have question for this feature 🤔
> 
> Since #323, the `chainQuery.blockQuery.blocks` query was introduced and it may have parameters like `desc`, `offset` and `limit`. IMO it seems the query of the libplanet-explorer can replace `topmostBlocks` query. 🤔 How do you think?
> 
I think you're right. The main reason for this PR was due to `9c-Launcher` using the `topmostBlocks` query to render the number of blocks mined on the launcher (Ref PR: https://github.com/planetarium/9c-launcher/pull/792). Since the most recent 500 blocks are prone to reorg, I thought it would be useful to have an `offset` parameters to lesson the players' confusion 😅.

I guess we have two options then:

1) Apply https://github.com/planetarium/NineChronicles.Headless/pull/542 & https://github.com/planetarium/9c-launcher/pull/792). 
2) Change `9c-Launcher` to use the `chainQuery` instead and close the 2 PR's above.

Could you let me know which option is better? 🙏 @moreal @riemannulus

moreal:
@area363 I didn't know the pull request for launcher was already ready. If this feature must be applied as soon as possible, the first option seems better. But if not, the second option seems the best way in my thought. 🤔

moreal (commented):
Could you rebase this pull request on the latest target branch? 🙏🏻

area363:
Branch is rebased! 🙇 @moreal

moreal (approved):

PR #586, opened 2021-07-27
Add option for determining behavior when the chain's tip is stale

limebell:
Used `string` instead of `enum` because `LibplanetNodeServiceProperties<T>` is generic class and it's very annoying.


ipdae:
is there any reason default value is "reboot"? It seems that "preload" worked by default before.

limebell:
> is there any reason default value is "reboot"? It seems that "preload" worked by default before.

Reboot option was default before rc100060. Preload option is newly added in this version.

longfin (approved):

ipdae (approved):

limebell:
I amended `render: true` parameter to `PreloadAsync()` which was missing.

longfin (approved):

PR #636, opened 2021-08-18
Tests involving crlf made OS independent

greymistcube:
Bunch of broken tests under Windows environment involving `\n` and `\r\n` fixed. Should now pass in both Linux and Windows.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,22 +3,28 @@ name: build
 on: [push, pull_request]
 
 jobs:
-  build-and-test:
-
-    runs-on: ${{ matrix.os }}
+  build-and-tests:
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
-
+        os:
+          - ubuntu-latest
+          - windows-latest
+    runs-on: ${{ matrix.os }}
     steps:
-    - uses: actions/checkout@v2
-      with:
-        submodules: recursive
-    - name: Setup .NET Core
-      uses: actions/setup-dotnet@v1
-      with:
-        dotnet-version: 3.1.301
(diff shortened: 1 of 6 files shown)


dahlia (commented):
Could you add a build job for Windows to the GitHub Actions workflow?

https://github.com/planetarium/NineChronicles.Headless/blob/047f9c6c1e2f99643f004a814a26f2b1c387df52/.github/workflows/build.yml#L8

See also <https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-running-with-multiple-operating-systems>.

area363:
The test doesn't seem to start due to the following issue:
```
Can't find any online and idle self-hosted runner in the current repository, account/organization that matches the required labels: 'ubuntu-latest , windows-latest'
Waiting for a self-hosted runner to pickup this job...
```

greymistcube:
Eh, I was lazy with `runs-on` argument. Thought I could get away with an array instead of using a matrix. Seems arrays are used for a different purpose.

Windows workflow build fails just like in the local environment:
```
2021-08-19T05:19:39.7560890Z   X NineChronicles.Headless.Executable.Tests.Commands.ChainCommandTest.Tip(storeType: RocksDb) [1s 63ms]
2021-08-19T05:19:39.7562343Z   Error Message:
2021-08-19T05:19:39.7563170Z    System.IO.IOException : The process cannot access the file 'LOCK' because it is being used by another process.
```

Linux equivalent:
```
2021-08-19T05:18:40.2896982Z   √ NineChronicles.Headless.Executable.Tests.Commands.ChainCommandTest.Tip(storeType: MonoRocksDb) [398ms]
```

I believe this is a separate issue tho.

PR #759, opened 2021-09-29
Bump lib9c

greymistcube:
Accommodate API changes in https://github.com/planetarium/lib9c/pull/640.

Note:
- Default mining behavior now checks if given miner address is allowed to mine according to the policy attached to a `BlockChain<T>` and does not attempt to mine and wait if not allowed.
- Due to removed `BlockPolicy` API, removed authorized miner check from headless.

----

- <https://github.com/planetarium/lib9c/pull/640>
- <https://github.com/planetarium/NineChronicles.Headless/pull/759>


dahlia:
@greymistcube I merged my PR #760, which depends on this PR, by mistake.  Could you open the same patch as a separate PR?

Issue #805, opened 2021-10-14
Re: #722 - GQL HnS once more broken


FioX0:
Hi team, 

sadly once more HnS got made a change against it and the gql equivalent isnt updated. when sending the gql query it accepts it but then flags an error 99.

As it is a reoccurring trend, may i suggest to ad some unit test around this, so you can easily identify if something broke.

thanks

area363:
Could you add a specific query that caused the query to fail? It works when I tried it so I'm wondering which arguments were used when it caused the error. Thanks!

kaila52:
How long will this issue be resolved?

PR #811, opened 2021-10-18
Add check playCount for `HackAndSlash` mutation

longfin:
(Note. This PR relies on https://github.com/planetarium/lib9c/pull/679)

It adds regression test for #805. 

Originally, I had tried to add a check about `txExecution`, but it was very difficult because the current testing fixture isn't realistic. so I changed the plan to add a check for just `HackAndSlash.playCount` 😢 .


area363:
Test fails 😢 https://github.com/planetarium/NineChronicles.Headless/pull/811/checks?check_run_id=3934996374#step:5:5299

PR #857, opened 2021-11-08
Unused arguments minimum difficulty and max transactions removed

greymistcube:
Should be fine since both variables aren't even referenced in `Program.cs`. 🙃


moreal (approved):
Nice catch! 👍🏻

greymistcube:
Bumps [lib9c] and removed ab additional argument `--authorized-miner`.

longfin:
It should be reflected in other infrastructure and 9c-launcher because it can occur `Unknown option` error. cc @planetarium/9c-backend @BasixKOR

PR #893, opened 2021-11-19
feat(cli): use default tx lifetime

moreal:
`Libplanet.Blockchain.Policies.VolatileStagePolicy` uses 180 minutes as default tx lifetime but the headless' option's default value was 60. This pull request makes 180 minutes, 3 hours as default value.


moreal:
@planetarium/9c-backend Is there other opinion about these changes?

area363 (approved):
I think increasing the default time is fine 👍

ipdae (approved):

PR #1020, opened 2022-01-20
Avoid NRE in GetStateBulk

ipdae:
- Set missing blockHash set.
- Avoid nre when value is null.

### DIFF
--- a/NineChronicles.Headless/BlockChainService.cs
+++ b/NineChronicles.Headless/BlockChainService.cs
@@ -119,10 +119,10 @@ namespace NineChronicles.Headless
             var hash = new BlockHash(blockHashBytes);
             var result = new Dictionary<byte[], byte[]>();
             Address[] addresses = addressBytesList.Select(b => new Address(b)).ToArray();
-            IReadOnlyList<IValue> values = _blockChain.GetStates(addresses);
+            IReadOnlyList<IValue> values = _blockChain.GetStates(addresses, hash);
             for (int i = 0; i < addresses.Length; i++)
             {
-                result.TryAdd(addresses[i].ToByteArray(), _codec.Encode(values[i]));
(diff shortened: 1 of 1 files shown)


area363 (approved):

NineChronicles.Headless/BlockChainService.cs
```diff
@@ -119,10 +119,10 @@ public UnaryResult<byte[]> GetState(byte[] addressBytes, byte[] blockHashBytes)
             var hash = new BlockHash(blockHashBytes);
             var result = new Dictionary<byte[], byte[]>();
             Address[] addresses = addressBytesList.Select(b => new Address(b)).ToArray();
-            IReadOnlyList<IValue> values = _blockChain.GetStates(addresses);
+            IReadOnlyList<IValue> values = _blockChain.GetStates(addresses, hash);
             for (int i = 0; i < addresses.Length; i++)
             {
-                result.TryAdd(addresses[i].ToByteArray(), _codec.Encode(values[i]));
+                result.TryAdd(addresses[i].ToByteArray(), _codec.Encode(values[i] ?? new Null()));
```
moreal, line 125: It is not a strong opinion, I think `Bencodex.Types.Null` and `null` have a different meaning and maybe there will be no `Null` states but if the case exists, clients can't distinguish them. So how about returning `null` instead of encoded `Bencodex.Types.Null` bytes. But it should check `null` can be sent through gRPC 🤔 

```suggestion
                result.TryAdd(addresses[i].ToByteArray(), values[i] is { } value ? _codec.Encode(value) : null);
```

ipdae: As far as I know, GetState is sending Bencodex.Null for the same reason.

dahlia: What @moreal pointed out above is valid, but AFAIK Nine Chronicles apparently does not depend on `null` states (or distinction between `null` and Bencodex `Null` at least) as of now.

PR #1038, opened 2022-02-04
Use `Server Garbage Collection` mode

moreal:
In .NET, there are two GC modes, *Workstation* and *Server*. You can see the difference between them at https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/workstation-server-gc.

This pull request makes the headless application to use Server GC mode.


NineChronicles.Headless.Executable/Program.cs
```diff
@@ -38,6 +39,7 @@ public class Program : CoconaLiteConsoleAppBase
 
         static async Task Main(string[] args)
         {
+            Console.WriteLine($"{nameof(GCSettings.IsServerGC)} = {GCSettings.IsServerGC}");
```
longfin, line 42: Do we still need this log?

longfin (approved):

NineChronicles.Headless.Executable/Program.cs
moreal: I dropped the commit.

longfin (approved):

PR #1049, opened 2022-02-10
E2E test

### LINKED ISSUE #956: Do E2E test
Currently, there are unit tests for almost GraphQL queries and commands or something. But there is no E2E now. So we cannot know that it is able to run really at end-user sight if we don't check manually for many cases. It should be automated.

moreal:
This pull request tries to resolve #956 issue. Currently, it doesn't check users can query GraphQL API and gRPC API. It tries to do E2E test at `ProgramTest.Run`.

### DIFF
--- a/NineChronicles.Headless.Executable.Tests/NineChronicles.Headless.Executable.Tests.csproj
+++ b/NineChronicles.Headless.Executable.Tests/NineChronicles.Headless.Executable.Tests.csproj
@@ -9,10 +9,12 @@
     </PropertyGroup>
 
     <ItemGroup>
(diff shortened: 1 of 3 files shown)


moreal:
Because 80 is a famous port, it seems used by other processes. I'll update it tomorrow and request reviews again. 🙏🏻

moreal:
You can review this pull request @planetarium/9c-backend, but the test will be failed because the `development` branch is broken since #1039 merged. It was fixed in `rc-v100102` but not here.

moreal:
Could you review this pull request? 🙏🏻 @planetarium/9c-backend

area363 (approved):

Issue #1032, opened 2022-01-27
GraphQL APIs doesn't work with `--rpc-http-server` option


moreal:
ASP.NET seems to apply only the last `ConfigureKerstel` setup. I reorder configurations, GraphQL and gRPC HTTP Gateway, then GraphQL works but gRPC HTTP Gateway doesn't.

moreal:
https://github.com/dotnet/aspnetcore/issues/17750#issuecomment-706237338

moreal:
Let me describe how we can turn on RPC API and GraphQL APIs:

To turn on RPC API, you have to use `--rpc-server` option with `--rpc-listen-port` with `--rpc-listen-host`. (`HostBuilderExtensions.UseNineChroniclesRPC()`)
And, to turn on GraphQL API, you have to use `--graphql-server`. (`GraphQLService`).

They use different methods to configure them. But `MapMagicOnionServices` is called in `GraphQLService` though I don't know its full context. 🤔 Maybe `GraphQLService` does all about HTTP services.

So the `--rpc-http-server` option should be also handled in `GraphQLService` and then it will use same port with GraphQL service and the `GraphQLService` should be renamed as `HttpService` or other.

In the perfect way, we should understand ASP.NET Core perfectly and should follow their convention.

moreal:
It is resolved by #1062

PR #1062, opened 2022-02-17
Correct `--rpc-http-server` option feature

moreal:
This pull request will close #1031, #1032.

It is not good practice to call the `UseKerstel` method twice times. It reset the previous state after calling it the second time. This pull request moves all initialization to `GraphQLService`.

And `--rpc-http-listen-host` and `--rpc-http-listen-port` are meaningless so this pull request removes them.

### DIFF
--- a/NineChronicles.Headless.Executable/Program.cs
+++ b/NineChronicles.Headless.Executable/Program.cs
@@ -182,8 +182,6 @@ namespace NineChronicles.Headless.Executable
             int txQuotaPerSigner = 10,
             bool rpcRemoteServer = false,
             bool rpcHttpServer = false,
-            string rpcHttpListenHost = "localhost",
-            int rpcHttpListenPort = 5000,
             [Option(Description = "The interval between block polling.  15 seconds by default.")]
             int pollInterval = 15,
             [Option(Description = "The maximum number of peers to poll blocks.  int.MaxValue by default.")]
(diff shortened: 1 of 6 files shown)


NineChronicles.Headless.Executable/Program.cs
```diff
@@ -22,6 +22,7 @@
 using System;
 using System.IO;
 using System.Threading.Tasks;
+using Org.BouncyCastle.Asn1.X509;
```
longfin, line 25: Is this `using` needed?

NineChronicles.Headless/Properties/GraphQLNodeServiceProperties.cs
```diff
@@ -1,3 +1,5 @@
+using Org.BouncyCastle.Asn1.X509;
```
longfin, line 1: It also seems unnecessary

moreal: Yes, I amended them because they are unnecessary. 🙏🏻

area363 (approved):

NineChronicles.Headless/HostBuilderExtensions.cs
```diff
@@ -93,45 +93,11 @@ RpcNodeServiceProperties properties
                 {
                     hostBuilder.ConfigureKestrel(options =>
                     {
-                        if (properties.HttpOptions is { } httpOptions)
-                        {
-                            options.ListenAnyIP(httpOptions.Port, listenOptions =>
-                            {
-                                listenOptions.Protocols = HttpProtocols.Http1;
-                            });   
-                        }
```
ipdae, line 102: i think we miss this option. is it tested?

moreal: I missed it but I didn't realize the point 🙏🏻 I tested it through accessing the `/swagger` endpoint. Maybe I guess that it is covered by the lines:

```csharp
options.ListenAnyIP((int)listenPort!, listenOptions =>
                        {
                            listenOptions.Protocols = HttpProtocols.Http1AndHttp2;
                        });
```

ipdae (approved):

Issue #968, opened 2021-12-24
GQL Query Request - Use Item via GQL (AP Potions)


FioX0:
Hi Team,

As previously advised on some meetings and on the discord, using AP Potions is extremely time consuming and inefficient.
To be able to use an AP Potion, you have to restart the game after each usage as the button then becomes greyed out.

This could be very simplified if there was a way to use said AP Potion via GQL, example:

Something along the lines of:

`mutation{
  action{
    UseItem(avatarAddress:"", itemId:"")
  }
}`

PR #1063, opened 2022-02-18
Tarog star update graphql actions

TarogStar:
To reduce transactions, those using the GraphQL endpoint can now add Play Count to boost their rewards.  The new Play Count variable is not required and defaults to 1.  Also added an action to Charge AP using GraphQL, which should close issue #968

### DIFF
--- a/NineChronicles.Headless/GraphTypes/ActionMutation.cs
+++ b/NineChronicles.Headless/GraphTypes/ActionMutation.cs
@@ -129,6 +129,12 @@ namespace NineChronicles.Headless.GraphTypes
                     {
                         Name = "consumableIds",
                         Description = "List of consumable id for use."
+                    }, 
+                    new QueryArgument<IntGraphType>
+                    {
+                        Name="playCount",
+                        DefaultValue = 1,
+                        Description = "Amount of times to boost rewards in one attempt."
                     }
                 ),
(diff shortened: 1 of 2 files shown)


longfin (approved):
LGTM! 

@planetarium/9c-backend Could you review it?

NineChronicles.Headless/GraphTypes/ActionMutation.cs
```diff
@@ -157,6 +164,7 @@ public ActionMutation(NineChroniclesNodeService service)
                             costumes = costumeIds,
                             equipments = equipmentIds,
                             foods = consumableIds,
+                            playCount = playCount < 1 ? 1 : playCount
```
moreal, line 167: In my thought, it'll be better to throw errors when unexpected negative numbers came than forcing to 1 implicitly.

area363 (approved):
👍

moreal (approved):
🚀

NineChronicles.Headless/GraphTypes/ActionMutation.cs
TarogStar: I was wondering whether to include that myself.  I adjusted the branch to remove that extra code.

longfin (approved):

PR #1071, opened 2022-02-22
Build Docker images for every tag

dahlia:
This makes the GitHub Actions to build & push a Docker image for each Git tag push, with a human-readable image tag name.  For example, if a tag named `v123456` is pushed, GitHub Actions will build and push the following tags and manifests:

- `planetariumhq/ninechronicles-headless:v123456-amd64` (tag)
- `planetariumhq/ninechronicles-headless:v123456-arm64v8` (tag)
- `planetariumhq/ninechronicles-headless:v123456` (manifest)

Note that `planetariumhq/ninechronicles-headless:git-*` tags will be still pushed too, but they are triggered by `branch` push events (not `tag` pushes).

### DIFF
--- a/.github/workflows/push_docker_image.yml
+++ b/.github/workflows/push_docker_image.yml
@@ -7,9 +7,15 @@ on:
       - rc-*
       - 9c-main
       - development
+    tags:
+      - v*
+
+env:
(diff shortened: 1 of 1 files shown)


area363 (approved):
🚀

moreal (approved):

PR #1133, opened 2022-03-17
Bump lib9c & Add store migration support

limebell:
Supports migration for RocksDB which removes column family. Please check https://github.com/planetarium/libplanet/pull/1862 for details.

### DIFF
--- a/Libplanet.Headless/Hosting/LibplanetNodeService.cs
+++ b/Libplanet.Headless/Hosting/LibplanetNodeService.cs
@@ -307,6 +307,14 @@ namespace Libplanet.Headless.Hosting
             {
                 try
                 {
+                    Log.Debug("Migrating RocksDB.");
+                    var chainPath = Path.Combine(path, "chain");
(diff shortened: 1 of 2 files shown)


Libplanet.Headless/Hosting/LibplanetNodeService.cs
```diff
@@ -313,6 +313,12 @@ public override async Task StopAsync(CancellationToken cancellationToken)
                         maxLogFileSize: 16 * 1024 * 1024,
                         keepLogFileNum: 1
                     );
+                    if (RocksDBStore.RocksDBStore.MigrateChainDBFromColumnFamilies(
```
dahlia, line 316: Is this call idempotent (in other words, okay if it's invoked more than once)?

limebell: Yes, it checks whether it is migrated first and does nothing if it is.

longfin (approved):

PR #1188, opened 2022-04-13
Fix weekly arena ordered infos

TarogStar:
GQL for weeklyArena query is no longer returning information on the rank and score of the players, this fixes that query to return information for the new storage format.


ipdae (approved):

longfin:
@TarogStar Thanks for the great work! sadly, [the build is failing due to a lack of API](https://github.com/planetarium/NineChronicles.Headless/runs/6002677716?check_suite_focus=true#step:5:796). could you check it?

TarogStar:
This doesn't actually return the same order as the game arena page.  Currently the arena sorts by Score and then by the deprecated CombatPoints, which were frozen some time back.  Should CombatPoints be calculated at run time?  It appears that they are for display in the client.  Is there another metric that should be used instead?

longfin:
> This doesn't actually return the same order as the game arena page. Currently the arena sorts by Score and then by the deprecated CombatPoints, which were frozen some time back. Should CombatPoints be calculated at run time? It appears that they are for display in the client. Is there another metric that should be used instead?

@ipdae Could you check it?

ipdae:
>This doesn't actually return the same order as the game arena page. Currently the arena sorts by Score and then by the deprecated CombatPoints, which were frozen some time back. Should CombatPoints be calculated at run time? It appears that they are for display in the client. Is there another metric that should be used instead?

@TarogStar Yes, currently order using only client side. there is no another metric.

longfin (approved):

PR #1191, opened 2022-04-13
Add default timeout to CheckPeer

TarogStar:
It appears that over time connections can get exhausted with unlimited timeouts.  With this small change my local node has had no problems for over 6 hours without restarting.


boscohyun (approved):

limebell:
Hi, it seems like the method `CheckPeer` is used in `HttpRequest` only but the endpoint is not called in [9c-launcher](https://github.com/planetarium/9c-launcher). You said the chain leads to performance enhancement, do you call the endpoint manually?

TarogStar:
This does seem like dead code, and not what affected my node.  It seems it may be affected by how many/what kind of transactions are flowing through the network at the time, rather than this.  It could also be related to me building it using VS 2022 in debug mode that is slightly more stable, but it is difficult to test as I don't know when the network will send tx that will cause issues.

PR #1192, opened 2022-04-14
Install curl

moreal:
Currently, it doesn't install the `curl` package because the base image already has it. But we always want to stay in the state prepared for .NET 6 except for only the target framework. In the .NET 6 image, it doesn't have the `curl` package so this pull request makes it install the package manually.


area363 (approved):
👍

PR #1206, opened 2022-04-21
feat(cli): print raw action into stdout as default

moreal:
This pull request makes the `action *` commands print the raw action into stdout, not file as default. If the file path is not given, it will print the raw action into stdout. So the compatibility will be stayed.


BasixKOR:
Great! This would allow simplifying the launcher implementation as well.

ipdae (approved):

PR #1272, opened 2022-05-19
Update Transaction API with actionQuery Compatibility

### LINKED ISSUE #1271: import planet key sign command

ipdae:
- resolve #1271 
- Import KeyCommand from Libplanet.
- Deprecate base64 using Transaction API.
- Update Transaction API with action query.

### DIFF
--- a/NineChronicles.Headless.Tests/GraphTypes/TransactionHeadlessQueryTest.cs
+++ b/NineChronicles.Headless.Tests/GraphTypes/TransactionHeadlessQueryTest.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
-using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using System.Security.Cryptography;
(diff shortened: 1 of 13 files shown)


NineChronicles.Headless.Tests/GraphTypes/TransactionHeadlessQueryTest.cs
```diff
@@ -316,8 +326,91 @@ private Task<ExecutionResult> ExecuteAsync(string query)
...
-                Store = _store
+                Store = _store,
+                NineChroniclesNodeService = _service
             });
-        }  
-    }     
+        }
+
+        [Fact]
+        public async Task SignTransaction()
```
moreal, line 335: It's not an important part. Is it intended as an E2E test, right?
Later, we should discuss the E2E tests convention (like all E2E should be named like `*ScenarioTest` or something..)

NineChronicles.Headless/GraphTypes/TransactionHeadlessQuery.cs
```diff
@@ -185,6 +181,133 @@ public TransactionHeadlessQuery(StandaloneContext standaloneContext)
...
+                        signature);
+                    signedTransaction.Validate();
+
+                    return signedTransaction.Serialize(true);
+                }
+            );
+
+            Field<NonNullGraphType<TxIdType>>(
+                name: "stageTx",
+                description: "Add a new transaction to staging and return TxId",
```
moreal, line 265: IMO, I suggest moving this query to mutation because I think it modifies server-side data (`BlockChain<T>`).

https://graphql.org/learn/queries/#mutations

> Most discussions of GraphQL focus on data fetching, but any complete data platform needs a way to modify server-side data as well.
>
> In REST, any request might end up causing some side-effects on the server, but by convention it's suggested that one doesn't use GET requests to modify data. GraphQL is similar - technically any query could be implemented to cause a data write. However, it's useful to establish a convention that any operations that cause writes should be sent explicitly via a mutation.

NineChronicles.Headless/GraphTypes/StandaloneMutation.cs
```diff
@@ -269,6 +270,55 @@ IConfiguration configuration
...
             );
+
+            Field<NonNullGraphType<TxIdType>>(
+                name: "stageTransaction",
+                description: "Add a new transaction to staging and return TxId",
+                arguments: new QueryArguments(
+                    new QueryArgument<NonNullGraphType<StringGraphType>>
+                    {
+                        Name = "payload",
+                        Description = "The hexadecimal of string for staging transaction."
```
moreal, line 281: Because `hexadecimal` is an adjective, `The hexadecimal string` seems more smooth.
```suggestion
                        Description = "The hexadecimal string of the transaction to stage."
```

moreal (approved):

PR #1278, opened 2022-05-23
Provides action commands related to `Stake`

moreal:
The next direction should be focused on the `actionQuery` GraphQL API, but this pull request provides action commands related to `Stake` now.


BasixKOR:
Some notes on why this PR was made: After looking around the implementation of `actionQuery` for a while I've come to the conclusion that this API would require a lot of changes to how the launcher creates and stages action, and I think the way how it would work with the launcher needs more consideration and potentially some changes on API.

ipdae (approved):

PR #1322, opened 2022-06-02
Add extra fields to battle arena

hskim881028:
related link : https://github.com/planetarium/lib9c/pull/1086


boscohyun (approved):
How about to remain the `RankingBattle` case?

PR #1374, opened 2022-06-20
feat(cli): introduce `--no-reduce-store` option

moreal:
This pull request introduces a new option, `--no-reduce-store` to disable `ReduceStore` which doesn't store all updated states except for updated fungible assets. There is a case to want to use all updated states for debugging or services. This pull request is for them.


ipdae (commented):
IMO, Renaming the option to `reducedStore` and keeping the default value false will helpful update convenience.

moreal:
> IMO, Renaming the option to `reducedStore` and keeping the default value false will helpful update convenience.

But if the default value is `false` and `--reduced-store` is required to use reduced store, previous users will run without reduced store. I'm not sure about it 🤔

moreal:
Is there no more opinions?

moreal:
Thanks for rebasing this branch 🙏🏻  @area363.

If there is no disagreement, I hope this pull request is merged.

area363 (approved):
🚀

PR #1405, opened 2022-07-01
Fix deploy gh pages workflow

ipdae:
https://github.com/planetarium/NineChronicles.Headless/actions/runs/2594765749


moreal (approved):
It looks good to me. If the test passed, it can be merged.

PR #1415, opened 2022-07-06
Championship arena GQL update

TarogStar:
I added some additional models, added the chain tip index to the StateContext, and inherited from IAccountStateDelta to allow all of the extension methods to get the arena data.  The old weekly arena is still available to get data for older arena rounds.

### DIFF
--- a/NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
+++ b/NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
@@ -54,7 +54,8 @@ namespace NineChronicles.Headless.GraphTypes
 
                     return new StateContext(
                         chain.ToAccountStateGetter(blockHash),
-                        chain.ToAccountBalanceGetter(blockHash)
+                        chain.ToAccountBalanceGetter(blockHash),
+                        chain.Tip.Index
                     );
                 }
             );
--- a/NineChronicles.Headless/GraphTypes/StateQuery.cs
+++ b/NineChronicles.Headless/GraphTypes/StateQuery.cs
@@ -8,6 +8,7 @@ using Libplanet;
 using Libplanet.Explorer.GraphTypes;
 using Nekoyume;
 using Nekoyume.Action;
+using Nekoyume.Model.Arena;
 using Nekoyume.Model.Item;
 using Nekoyume.Model.State;
 using Nekoyume.TableData;
@@ -58,7 +59,7 @@ namespace NineChronicles.Headless.GraphTypes
                     var index = context.GetArgument<int>("index");
                     if (context.Source.GetState(RankingState.Derive(index)) is { } state)
                     {
-                        return new RankingMapState((Dictionary) state);
+                        return new RankingMapState((Dictionary)state);
                     }
 
                     return null;
@@ -68,7 +69,7 @@ namespace NineChronicles.Headless.GraphTypes
                 description: "State for shop.",
                 deprecationReason: "Shop is migrated to ShardedShop and not using now. Use shardedShop() instead.",
                 resolve: context => context.Source.GetState(Addresses.Shop) is { } state
-                    ? new ShopState((Dictionary) state)
+                    ? new ShopState((Dictionary)state)
                     : null);
             Field<ShardedShopStateV2Type>(
                 name: "shardedShop",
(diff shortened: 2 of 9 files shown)


NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
```diff
@@ -54,7 +54,8 @@ public StandaloneQuery(StandaloneContext standaloneContext, IConfiguration confi
 
                     return new StateContext(
                         chain.ToAccountStateGetter(blockHash),
-                        chain.ToAccountBalanceGetter(blockHash)
+                        chain.ToAccountBalanceGetter(blockHash),
+                        chain.Tip.Index
```
ipdae, line 58: Even when a block hash is given as an argument, but block index always use latest index. it may cause confusion.

NineChronicles.Headless/GraphTypes/States/StateContext.cs
```diff
@@ -1,21 +1,30 @@
...
 {
-    public class StateContext
+    public class StateContext : IAccountStateDelta
     {
-        public StateContext(AccountStateGetter accountStateGetter, AccountBalanceGetter accountBalanceGetter)
+        public StateContext(AccountStateGetter accountStateGetter, AccountBalanceGetter accountBalanceGetter, long chainTipIndex)
         {
             AccountStateGetter = accountStateGetter;
             AccountBalanceGetter = accountBalanceGetter;
+            ChainTipIndex = chainTipIndex;
```
ipdae, line 16: Where is chainTipIndex used? I can't see where it's used, is there a reason it was added?

ipdae (commented):
Thanks for your contribution. I have a few questions, so I left a comment.

NineChronicles.Headless/GraphTypes/StateQuery.cs
```diff
@@ -91,9 +92,80 @@ public StateQuery()
...
+                            continue;
+                        }
+
+                        var arenaInfo = new ChampionArenaInfo();
+                        arenaInfo.AvatarAddress = participant;
+                        arenaInfo.Win = arenaInformation.Win;
+                        arenaInfo.Ticket = arenaInformation.Ticket;
+                        arenaInfo.Lose = arenaInformation.Lose;
+                        arenaInfo.Score = arenaScore.Score;
+                        arenaInformations.Add(arenaInfo);
```
boscohyun, line 165: If you want to sorting this `arenaInformations` same with client, and check [this function][1].

- Ranking is determined by score.
- For duplicate ranks, the same rank is displayed.
   - In this case, the same rank is a rank that is separated by the number of duplicates from the previous rank.
- Duplicate Rank The next rank is displayed as a rank separated by the number of duplicates.

e.g., You can check what we expect with [Data source][2] and [Unit test][3]

And we will move this ranking function from [9c][4] to [lib9c][5]


[1]: https://github.com/planetarium/NineChronicles/blob/main/nekoyume/Assets/_Scripts/State/RxProps.Arena.cs#L452-L526
[2]: https://github.com/planetarium/NineChronicles/blob/main/nekoyume/Assets/Tests/EditMode/State/RxPropsTest.cs#L10-L38
[3]: https://github.com/planetarium/NineChronicles/blob/main/nekoyume/Assets/Tests/EditMode/State/RxPropsTest.cs#L10-L38
[4]: https://github.com/planetarium/NineChronicles
[5]: https://github.com/planetarium/lib9c

NineChronicles.Headless/GraphTypes/States/StateContext.cs
```diff
@@ -1,21 +1,30 @@
...
+using System.Collections.Immutable;
 using Bencodex.Types;
 using Libplanet;
 using Libplanet.Action;
 using Libplanet.Assets;
 
 namespace NineChronicles.Headless.GraphTypes.States
 {
-    public class StateContext
+    public class StateContext : IAccountStateDelta
```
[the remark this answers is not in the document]
TarogStar: IAccountStateDelta allows using GetSheets, TryGetArenaParticipants, TryGetArenaInformation, TryGetArenaScore  Would it make sense to just import those methods and duplicate them in the StateContext?

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
TarogStar: I submitted a new commit to address the issue to use the right block index.

NineChronicles.Headless/GraphTypes/States/StateContext.cs
TarogStar: It looks like I didn't get everything over to this branch previously.  I needed the index to determine if 0 tickets is actually 0 or should be 8 and they haven't battled since the reset.  I have now included the code that uses the index.

PR #1454, opened 2022-08-01
Add `prune-states` command

area363:
This PR adds a `chain command` that prunes the `states` of the 9c store (similar to `NineChronicles.Snapshot`).

With this command, downloading a new snapshot when the `states` size is too large is no longer needed.

One issue is that this command takes close to `30 mins` to complete in our current 9c store, which has an initial `states` size of about `4GB`. I'll add a more detailed issue on the `libplanet` repo.

### DIFF
--- a/NineChronicles.Headless.Executable.Tests/Commands/ChainCommandTest.cs
+++ b/NineChronicles.Headless.Executable.Tests/Commands/ChainCommandTest.cs
@@ -1,8 +1,11 @@
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.IO;
 using System.Linq;
 using System.Security.Cryptography;
+using System.Text;
+using System.Text.Json;
 using System.Threading.Tasks;
 using Libplanet;
 using Libplanet.Action;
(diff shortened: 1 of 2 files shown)


area363:
@moreal I discussed your suggestion to put in the snapshot function with @longfin and I think we can merge this command first and use it as a subcommand to create a separate snapshot command. How does that sound?

moreal:
> I think we can merge this command first and use it as a subcommand to create a separate snapshot command. How does that sound?

@area363 Then, it means that `headless chain prune-states` will be introduced first and it will be moved like `headless snapshot prune-states`, right?

area363:
> > I think we can merge this command first and use it as a subcommand to create a separate snapshot command. How does that sound?
> 
> @area363 Then, it means that `headless chain prune-states` will be introduced first and it will be moved like `headless snapshot prune-states`, right?

Yes, I think so 👍

moreal:
> Yes, I think so 👍

Then it sounds good to me 👍🏻

area363:
Ok, i'll merge this PR for now. Thanks for your input! 🙇

PR #1479, opened 2022-08-12
Introduce world boss models

ipdae:
- AddressQuery for find derive addresses
- RaiderState
- WorldBossState
- WorldBossKillRecord

### DIFF
--- a/NineChronicles.Headless/GraphTypes/StateQuery.cs
+++ b/NineChronicles.Headless/GraphTypes/StateQuery.cs
@@ -8,6 +8,7 @@ using Libplanet;
 using Libplanet.Explorer.GraphTypes;
 using Nekoyume;
 using Nekoyume.Action;
+using Nekoyume.Extensions;
 using Nekoyume.Model.Item;
 using Nekoyume.Model.State;
 using Nekoyume.TableData;
@@ -318,6 +319,51 @@ namespace NineChronicles.Headless.GraphTypes
                     }
 
                     return null;
+                }
+            );
+
+            Field<RaiderStateType>(
+                name: "raiderState",
+                description: "world boss season information.",
+                arguments: new QueryArguments(
+                    new QueryArgument<NonNullGraphType<AddressType>>
+                    {
+                        Name = "raiderAddress",
+                        Description = "address of world boss season."
+                    }
+                ),
+                resolve: context =>
+                {
+                    var raiderAddress = context.GetArgument<Address>("raiderAddress");
+                    IReadOnlyList<IValue?> values = context.Source.AccountStateGetter(new[] { raiderAddress });
+                    if (values[0] is List list)
+                    {
+                        return new RaiderState(list);
+                    }
+
+                    return null;
+                }
+            );
+
+            Field<NonNullGraphType<IntGraphType>>(
+                "raidId",
+                description: "world boss season id by block index.",
+                arguments: new QueryArguments(new QueryArgument<LongGraphType>
+                {
+                    Name = "blockIndex"
+                }),
+                resolve: context =>
+                {
+                    var blockIndex = context.GetArgument<long>("blockIndex");
+                    var sheet = new WorldBossListSheet();
+                    var address = Addresses.GetSheetAddress<WorldBossListSheet>();
+                    IReadOnlyList<IValue?> values = context.Source.AccountStateGetter(new[] {address});
+                    if (values[0] is Text text)
+                    {
+                        sheet.Set(text);
+                    }
+
+                    return sheet.FindRaidIdByBlockIndex(blockIndex);
                 }
             );
         }
--- a/NineChronicles.Headless/GraphTypes/StateQuery.cs
+++ b/NineChronicles.Headless/GraphTypes/StateQuery.cs
@@ -324,7 +324,7 @@ namespace NineChronicles.Headless.GraphTypes
 
             Field<RaiderStateType>(
                 name: "raiderState",
-                description: "world boss season information.",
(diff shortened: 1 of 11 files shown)


NineChronicles.Headless/GraphTypes/States/WorldBossKillRewardRecordType.cs
```diff
@@ -0,0 +1,28 @@
...
+                    description: "check reward already claimed. if already claimed return true.",
+                    resolve: context => context.Source.Value);
+            }
+        }
+
+        public WorldBossKillRewardRecordType()
+        {
+            Field<ListGraphType<WorldBossKillRewardRecordMapType>>(
+                "map",
+                resolve: context => context.Source);
```
moreal, line 25: The map and those elements seems non-nullable. If there is no records, it will return empty list `{ map: [] }`.

moreal (commented):
I left these reviews for GraphQL with looking world boss models but I'm not expert for this world boss feature.

moreal:
Oops, I'm sorry for my mistake about this review 😞. I missed this pull request is based on #1477 so the comments that should be located at #1477, are also here.

ipdae:
@moreal can I close #1477 and fix with it here?

moreal:
> @moreal can I close #1477 and fix with it here?

@ipdae Sure, if you want that 👍🏻

NineChronicles.Headless/GraphTypes/StateQuery.cs
```diff
@@ -320,6 +321,86 @@ public StateQuery()
...
+                {
+                    var blockIndex = context.GetArgument<long>("blockIndex");
+                    var sheet = new WorldBossListSheet();
+                    var address = Addresses.GetSheetAddress<WorldBossListSheet>();
+                    if (context.Source.GetState(address) is Text text)
+                    {
+                        sheet.Set(text);
+                    }
+
+                    return sheet.FindRaidIdByBlockIndex(blockIndex);
```
moreal, line 364: Though I'm not an expert of the world boss feature, is it okay to find raid id even if there is no sheet? (empty sheet).

When I saw the `FindRaidIdByBlockIndex(long)` extension method, it seems failed when there is no elements:

```csharp
        public static WorldBossListSheet.Row FindRowByBlockIndex(this WorldBossListSheet sheet,
            long blockIndex)
        {
            return sheet.OrderedList
                .First(r =>
                    r.StartedBlockIndex <= blockIndex &&
                    blockIndex <= r.EndedBlockIndex
                );
        }
```

Maybe the type of `raidId` should be nullable `IntGraphType`.

NineChronicles.Headless/GraphTypes/States/WorldBossKillRewardRecordType.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace NineChronicles.Headless.GraphTypes.States
+{
+    public class WorldBossKillRewardRecordType : ObjectGraphType<WorldBossKillRewardRecord>
+    {
+        private class WorldBossKillRewardRecordMapType : ObjectGraphType<KeyValuePair<int, bool>>
+        {
+            public WorldBossKillRewardRecordMapType()
+            {
+                Field<IntGraphType>("bossLevel", resolve: context => context.Source.Key);
```
moreal, line 13:
```suggestion
                Field<NonNullableGraphType<IntGraphType>>("bossLevel", resolve: context => context.Source.Key);
```

NineChronicles.Headless/GraphTypes/States/WorldBossKillRewardRecordType.cs
```diff
@@ -0,0 +1,28 @@
...
+                Field<NonNullGraphType<BooleanGraphType>>(
+                    "claimed",
+                    description: "check reward already claimed. if already claimed return true.",
+                    resolve: context => context.Source.Value);
+            }
+        }
+
+        public WorldBossKillRewardRecordType()
+        {
+            Field<NonNullGraphType<ListGraphType<WorldBossKillRewardRecordMapType>>>(
```
moreal, line 23: I think the element's type is also non-nullable because it doesn't return like:

```
{
  map: [
    {
      bossLevel: 0,
      claimed: false
    },
    null,
    {
      bossLevel: 3
      claimed: true
    }
  ]
}
```

```suggestion
            Field<NonNullGraphType<ListGraphType<NonNullGraphType<WorldBossKillRewardRecordMapType>>>>(
```

moreal (commented):
Thank you for applying the reviews 🙏🏻 
I left three more comments which I missed 🙇🏻‍♂️

NineChronicles.Headless/GraphTypes/StateQuery.cs
ipdae: `FindRowByBlockIndex` throw exception when sheet is empty. so the type seems to be non-nullable. I think it's like `Linq.First` and `Linq.FirstorDefault`.

moreal: Oops you're right. I'll resolve this comment.

---

I understood that with the below example:

```csv
id,boss_id,started_block_index,ended_block_index,fee,ticket_price,additional_ticket_price,max_purchase_count
1,900001,0,100,1,5,5,40
2,900001,200,300,1,5,5,40
```

Throwing the error from `Linq.First` means there is no boss for the block index. For example, it will also throw error when the block index is more than 300 with the sheet. If the sheet state doesn't exist, it works like the example.

moreal (approved):
🚀

PR #1488, opened 2022-08-16
Make GenesisCommandTest passes: resolve #1487

### LINKED ISSUE #1487: Failing `Mine` Test in `GenesisCommandTest`
It appears that the CI test for `Mine` in `GenesisCommandTest`  is failing after https://github.com/planetarium/NineChronicles.Headless/pull/1470.

Is there any plan to fix this issue? 🙏 @U-lis cc @moreal

U-lis:
This PR makes `GenesisCommandTest` passes.

### Changelist
- Apply new genesis config structure to unittest
- Only `DataConfig` is required, so change expected exception result of one testcase.

Sorry for breaking integrities 🙏

### DIFF
--- a/NineChronicles.Headless.Executable.Tests/Commands/GenesisCommandTest.cs
+++ b/NineChronicles.Headless.Executable.Tests/Commands/GenesisCommandTest.cs
@@ -6,6 +6,7 @@ using Cocona;
 using Libplanet;
 using Libplanet.Blocks;
 using Libplanet.Crypto;
+using Nekoyume.Action;
 using NineChronicles.Headless.Executable.Commands;
 using NineChronicles.Headless.Executable.Tests.IO;
 using Xunit;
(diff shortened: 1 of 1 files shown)


NineChronicles.Headless.Executable.Tests/Commands/GenesisCommandTest.cs
```diff
@@ -62,43 +126,25 @@ public void Mine(bool tablePathExist, bool goldDistributionPathExist, bool actio
...
-                        "0000000000000000000000000000000000000002",
-                        "0000000000000000000000000000000000000003",
-                        "0000000000000000000000000000000000000004"
-                    }
+                    "0000000000000000000000000000000000000001",
+                    "0000000000000000000000000000000000000002",
+                    "0000000000000000000000000000000000000003",
+                    "0000000000000000000000000000000000000004"
                 }
             };
```
moreal, line 146: Is there a reason to stay the deprecated config also?

U-lis:
@moreal  for both comments:
I leaved those deprecated configs for historical reason only.
I thought that "Someday we may need this again... Commit history has everything but history is not esay to reach." 🤔 

Leaving old code has both ups and downs and I like leaving history for first touch and delete it when someone who edits the same code again. << lt's just my style
If there is policy to manage deprecated/unused code, I'll follow the rule with happy :-)

moreal:
> I leaved those deprecated configs for historical reason only. I thought that "Someday we may need this again... Commit history has everything but history is not esay to reach." 🤔
> 
> Leaving old code has both ups and downs and I like leaving history for first touch and delete it when someone who edits the same code again. << lt's just my style If there is policy to manage deprecated/unused code, I'll follow the rule with happy :-)

Thank you for describing your opinion 🙇🏻‍♂️ 

> Someday we may need this again... Commit history has everything but history is not easy to reach.

I also think we may need this again someday. But I wonder whether the stayed lines can provide the useful information when new contributors (even if in-company people) tries to recover that. I think the person to work on that, should find the history to undertstand the feature (like PR or commits histories) and can revert with the specific commit. (e.g. `git revert <COMMIT>`)

> Leaving old code has both ups and downs and I like leaving history for first touch and delete it when someone who edits the same code again. << lt's just my style If there is policy to manage deprecated/unused code, I'll follow the rule with happy :-)

About the above line, I wrote my thought on the above section. I don't think there is a specified document about the coding guideline and my thought I wrote may be personal. About this, I think we should listen to opinions other reviewers' thoughts cc @longfin @area363.

U-lis:
Any progress here?
@moreal 
I think the best way from now is to make another commit to delete old codes.
And someday we can revert or reference that commit when we need.

moreal:
> Any progress here? @moreal I think the best way from now is to make another commit to delete old codes. And someday we can revert or reference that commit when we need.

I think people cannot review this pull request because of some reason (e.g. busy). I agree with your opinion.

moreal (approved):

PR #1522, opened 2022-08-29
Rc v100282 to development

Atralupus:


moreal (commented):
Maybe it'll be passed when the Lib9c was bumped as the resolved development (including Libplanet bumps)

moreal (approved):

PR #1560, opened 2022-09-14
Add actionTxQuery field

### LINKED ISSUE #958: Support `nonce` argument for `transaction.createUnsignedTx` GraphQL query
Currently, there is a `transaction.createUnsignedTx` query to make an unsigned transaction with a public key and action. But there is no `nonce` argument to configure the `nonce` of the transaction manually. It should be provided.

BasixKOR:
Resolves #958.


BasixKOR:
Maybe we could pass the arguments to ActionQuery instead of poking the parent around

moreal:
@BasixKOR I added simple three commits, could you check it?

BasixKOR:
The added tests look great to me. Feel free to make changes as you see fit!

moreal (approved):
Though I worked also, I think this pull request can be merged ✅

PR #1568, opened 2022-09-16
feat(graphql): deprecate GraphQL APIs to handle private key

moreal:
This pull request is related with #1567 discussion.

I think we should recommend to use `planet key` command to handle private key. This pull request deprecates the GraphQL fields and types:

 - `KeyStoreType` type
 - `KeyStoreMutation` type
 - `mutation.keyStore` field
 - `query.keyStore` field


area363 (approved):
👍

PR #1572, opened 2022-09-19
chore: support bls public key in parser

limebell:
Support argument input for BLS public key

### DIFF
--- a/Libplanet.Headless/PropertyParser.cs
+++ b/Libplanet.Headless/PropertyParser.cs
@@ -31,7 +31,7 @@ namespace Libplanet.Headless
                     $"--peer '{peerInfo}', should have format <pubkey>,<host>,<port>");
             }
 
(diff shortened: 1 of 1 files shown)


Libplanet.Headless/PropertyParser.cs
```diff
@@ -52,7 +52,7 @@ public static BoundPeer ParsePeer(string peerInfo)
             catch (Exception e)
             {
                 throw new PeerInvalidException(
-                    $"--peer '{peerInfo}' seems invalid.\n" +
+                    $"Given peer (-peer or --consensus-peer) '{peerInfo}' seems invalid.\n" +
```
greymistcube, line 55:
```suggestion
                    $"Given peer (--peer or --consensus-peer) '{peerInfo}' seems invalid.\n" +
```

Issue #1556, opened 2022-09-13
Run headless server with config file, not cli params


U-lis:
Currently, the only way to run headless server is type all the cli params when run headless server.
To make running server easier, we should merge this params to settings file like `config.json` or `appsettings.json`

U-lis:
### Work plan

1. Add a section in `appsettings.json` to store headless parameters.
2. Change all headless parameters to `optional`
3. `appsettings.json` has default headless parameters
4. Optional parameters will overwrite settings from `appsettings.json`

### Considerations
- `appsettings.json` sould be independent from code/package itself.
    - `appsettings.json` path should be passed through parameter.
- Legacy running (using longlong params) must wotk perfectly just like as-is.

PR #1577, opened 2022-09-21
Use appsettings file to run headless server

U-lis:
## Summary
Now we can use `appsettings.json` to run headless.
`appsettings.json` can be local file or web storage like S3.

## Problem
We have to pass a bunch of command line options to run headless server so far and this is very inconvenient work.

## Solution
Read headless launch options from `appsettings.json`.
The old CLI options are still alive and this will overwrite options from `appsettings.json`

---

Related issue: https://github.com/planetarium/NineChronicles.Headless/issues/1556

### DIFF
--- a/NineChronicles.Headless.Executable/Configuration.cs
+++ b/NineChronicles.Headless.Executable/Configuration.cs
@@ -1,0 +1,245 @@
+using NineChronicles.Headless.Properties;
+
+namespace NineChronicles.Headless.Executable
+{
+    public class Configuration
+    {
+        public const bool DefaultNoReduceStore = false;
+        public const bool DefaultNoMiner = false;
+        public const int DefaultWorkers = 5;
+        public const int DefaultConfirmations = 0;
+        public const bool DefaultNonblockRenderer = false;
+        public const int DefaultNonblockRendererQueue = 512;
+        public const int DefaultMessageTimeout = 60;
+        public const int DefaultTipTimeout = 60;
+        public const int DefaultDemandBuffer = 1150;
+        public const bool DefaultSkipPreload = true;
+        public const int DefaultMinimumBroadcastTarget = 10;
+        public const int DefaultBucketSize = 16;
+        public const string DefaultChainTipStaleBehaviorType = "reboot";
+        public const int DefaultMaximumPollPeers = int.MaxValue;
+
+        public const bool DefaultGraphQLServer = false;
+        public const bool DefaultNoCors = false;
+        public const bool DefaultRpcServer = false;
+        public const string DefaultRpcListenHost = "0.0.0.0";
+
+        public class DevConfiguration
+        {
+            public int? BlockInterval { get; set; }
+            public int? ReorgInterval { get; set; }
+        }
+
+        public string? AppProtocolVersionString
+        {
+            get;
+            set;
+            // set => AppProtocolVersion = Libplanet.Net.AppProtocolVersion.FromToken(value);
+        }
+
+        public AppProtocolVersion? AppProtocolVersion { get; set; }
+
+        public string[]? TrustedAppProtocolVersionSignerStrings
+        {
+            get;
+            set;
+            // set => TrustedAppProtocolVersionSigners = value
+            // ?.Select(s => new PublicKey(ByteUtil.ParseHex(s)))
+            // ?.ToHashSet();
+        }
+
+        public HashSet<PublicKey>? TrustedAppProtocolVersionSigners { get; set; }
+        public string? GenesisBlockPath { get; set; }
+        public string? Host { get; set; }
+        public ushort? Port { get; set; }
+
+        public string? SwarmPrivateKeyString
+        {
+            get;
+            set;
+            // set => SwarmPrivateKey = string.IsNullOrEmpty(value)
+            // ? new PrivateKey()
+            // : new PrivateKey(ByteUtil.ParseHex(value));
+        }
+
+        public PrivateKey? SwarmPrivateKey { get; set; }
+        public int? Workers { get; set; }
+
+        // Storage
+        public string? StoreType { get; set; }
+        public string? StorePath { get; set; }
+        public bool? NoReduceStore { get; set; }
+
+        // Miner
+        public bool? NoMiner { get; set; }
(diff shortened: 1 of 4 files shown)


moreal (approved):

NineChronicles.Headless.Executable/appsettings.json
```diff
@@ -84,5 +88,38 @@
                 }
             }
         ]
+    },
+    "Headless": {
+        "AppProtocolVersionString": "100291/6ec8E598962F1f475504F82fD5bF3410eAE58B9B/MEUCIQCA3lzUAt0QBfG.+ezw4CQ69zBy669sANEt5juSJgzqcgIgbozfpcyeuKJDeJoT5exyGYDYBqCpxklsMEfs0SQ6qzo=/ZHUxNjpXaW5kb3dzQmluYXJ5VXJsdTcyOmh0dHBzOi8vcmVsZWFzZS5uaW5lLWNocm9uaWNsZXMuY29tL21haW4vdjEwMDI5MS9sYXVuY2hlci92MS9XaW5kb3dzLnppcHU5OnRpbWVzdGFtcHUxMDoyMDIyLTA5LTA3ZQ==",
+        "GenesisBlockPath": "https://release.nine-chronicles.com/genesis-block-9c-main",
+        "StoreType": "rocksdb",
+        "StorePath": "/home/ulismoon/AppData/Local/planetarium/9c-main-partition",
```
moreal, line 96: This should be replaced with generic location. In geth case, it stores data in `~/.ethereum`

moreal (changes requested):

NineChronicles.Headless.Executable/appsettings.json
U-lis: Oh I missed to change that thing after local test :sob:

ipdae:
we can divide command arguments and `HeadlessConfig`, and set default value can be remove nullable argument in `HeadlessConfig`

NineChronicles.Headless.Executable/appsettings.json
```diff
@@ -84,5 +88,38 @@
                 }
             }
         ]
+    },
+    "Headless": {
+        "AppProtocolVersionString": "100291/6ec8E598962F1f475504F82fD5bF3410eAE58B9B/MEUCIQCA3lzUAt0QBfG.+ezw4CQ69zBy669sANEt5juSJgzqcgIgbozfpcyeuKJDeJoT5exyGYDYBqCpxklsMEfs0SQ6qzo=/ZHUxNjpXaW5kb3dzQmluYXJ5VXJsdTcyOmh0dHBzOi8vcmVsZWFzZS5uaW5lLWNocm9uaWNsZXMuY29tL21haW4vdjEwMDI5MS9sYXVuY2hlci92MS9XaW5kb3dzLnppcHU5OnRpbWVzdGFtcHUxMDoyMDIyLTA5LTA3ZQ==",
+        "GenesisBlockPath": "https://release.nine-chronicles.com/genesis-block-9c-main",
+        "StoreType": "rocksdb",
+        "StorePath": "~/AppData/Local/planetarium/9c-main-partition",
```
moreal, line 96: Is this path the default path used by 9c-launcher? (because 9c officially supports only Windows platform now)

moreal (commented):
Good works!

NineChronicles.Headless.Executable/appsettings.json
BasixKOR: Something I'm slightly worried is that `~` is a POSIX convention and may not work on .NET or Windows.

BasixKOR, line 96: Something I'm slightly worried about is that `~` is a POSIX convention and may not work on .NET or Windows.

U-lis: > Is this path the default path used by 9c-launcher? (because 9c officially supports only Windows platform now)

The launcher uses `[HOME]/AppData/Local/~~~` and finds `HOME` using [this code](https://github.com/planetarium/9c-launcher/blob/main/src/config.ts#L161-L166)
I think just put placeholder like `[path/to/9c-main-partition]` is kind of feasible but cannot sure :thinking:

U-lis: It's little bit tricky. In this case I prefer to put placeholder like `[path/to/9c-main-partition]` but this `appsetting.json` is not an example so placeholder is not good choice to pick.
But there is no silver bullet to cover everything...

moreal: Hmm, then how about make `appsettings.json`'s `StorePath` make nullable and implement the default vlaue in the `Configuration.StorePath` property 🤔 .

U-lis: @moreal  That makes sense. :+1:

NineChronicles.Headless.Executable/Configuration.cs
```diff
@@ -0,0 +1,213 @@
...
+        public string? StoreType { get; set; }
+
+        public string? StorePath { get; set; } =
+            Path.Combine(
+                new string[]
+                {
+                    Environment.GetFolderPath(Environment.SpecialFolder.UserProfile),
+                    "AppData", "Local", "planetarium", "9c-main-partition"
+                }
+            );
```
moreal, line 37: It seems good 👍🏻 But if you intended to target the 9c-launcher's store path, how about branching the store path by OS platform like 9c-launcher does?

https://github.com/planetarium/9c-launcher/blob/377138073823f1f6efd9bdfc3466fe47aa9ee4df/src/config.ts#L161-L166 

It seems it can use like https://learn.microsoft.com/en-us/dotnet/api/system.operatingsystem.ismacos?view=net-6.0#system-operatingsystem-ismacos 🤔 

Of course, it officially supports Windows yet so it can be passed now 🤔

moreal (approved):

NineChronicles.Headless.Executable/Configuration.cs
BasixKOR: This may help: https://developers.redhat.com/blog/2018/11/07/dotnet-special-folder-api-linux

PR #1602, opened 2022-09-28
Feature/commands from lib9c tools

U-lis:
Commands from [Lib9c.Tools](https://github.com/planetarium/lib9c/tree/development/.Lib9c.Tools) to `NineChronicles.Headless`

### DIFF
--- a/NineChronicles.Headless.Executable/Commands/MarketCommand.cs
+++ b/NineChronicles.Headless.Executable/Commands/MarketCommand.cs
@@ -1,0 +1,226 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Bencodex;
+using Bencodex.Types;
+using Cocona;
+using Lib9c.DevExtensions;
+using Lib9c.Model.Order;
+using Libplanet;
+using Libplanet.Assets;
+using Libplanet.Blockchain;
+using Libplanet.Blocks;
+using Libplanet.Store;
+using Libplanet.Tx;
+using Nekoyume.Action;
+using Nekoyume.Model.Item;
+using NineChronicles.Headless.Executable.IO;
+using Serilog.Core;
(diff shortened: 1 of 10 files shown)


U-lis:
Mistake: click `merge` instead of `rebase` :sob:

moreal (commented):
It isn't important in this pull request the action commands in the tx command should be removed.

NineChronicles.Headless.Executable/Commands/MarketCommand.cs
```diff
@@ -0,0 +1,226 @@
...
+        public MarketCommand(IConsole console)
+        {
+            _console = console;
+        }
+
+        [Command(Description = "Query market transactions.")]
+        public void Query(
+            [Option('v', Description = "Print more logs.")]
+            bool verbose,
+            [Option('s', Description = "Path to the chain store.")]
```
moreal, line 38: Is it okay to fix the store type as rocksdb?

moreal (approved):

NineChronicles.Headless.Executable/Commands/MarketCommand.cs
U-lis: > Is it okay to fix the store type as rocksdb?

Hmm.. I didn't think about that deeply :thinking: 
With my gut feeling, there is no practical scenario to query market info to the chain running with volatile storage.

And this function uses `Lib9c.DevExtensions.Utils.GetBlockChain()` and this assumes storage type to RocksDB.
I'll try to change this to support multiple StoreType.

moreal: Ah okay, I don't feel to need to fix it in this pull request. Because this pull request's purpose was to take the commands from lib9c-tools. So I approved and you can merge this pull request 😉

PR #1644, opened 2022-10-20
Merge rc-v100310 to development

OnedgeLee:
SSIA


moreal:
In my thought, the `from`(maybe) branch should be `main` instead of `rc-v100302` to include also `v100302` tag commit and missed hotfix commits.

PR #1663, opened 2022-10-28
Easy Gql transfers

TarogStar:
Allow transfers to be done through GraphQl much easier, and allow choosing crystal as well.  People are currently using complex batch scripts which could be corrupted or cause problems if used incorrectly or altered by a malicious party.

### DIFF
--- a/NineChronicles.Headless/GraphTypes/StandaloneMutation.cs
+++ b/NineChronicles.Headless/GraphTypes/StandaloneMutation.cs
@@ -10,6 +10,7 @@ using Libplanet.Explorer.GraphTypes;
 using Libplanet.Tx;
 using Microsoft.Extensions.Configuration;
 using Nekoyume.Action;
+using Nekoyume.Helper;
 using Nekoyume.Model.State;
 using Serilog;
 using System;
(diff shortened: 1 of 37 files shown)


NineChronicles.Headless/GraphTypes/StandaloneMutation.cs
```diff
@@ -154,17 +155,10 @@ IConfiguration configuration
                         Description = "A string value of the value to be transferred.",
                         Name = "amount",
                     },
-                    new QueryArgument<NonNullGraphType<LongGraphType>>
+                    new QueryArgument<NonNullGraphType<CurrencyEnumType>>
                     {
-                        Description = "A sender's transaction counter. You can get it through nextTxNonce().",
-                        Name = "txNonce",
-                    },
```
moreal, line 161: I remembered it was designed to restrict mistake to send multiple times. cc @longfin

longfin:
@TarogStar Thanks for your commitment and sorry for the late reply. I think that there're two points we should consider.

1. We want to reduce the usage of `StandaloneMutation` except `stageTransaction()` since all other mutations require a private key in any form (i.e., `.MinerKey` or the parameter of mutations). I believe the combination of [`actionTxQuery`](https://github.com/planetarium/NineChronicles.Headless/pull/1560/f) + ECDSA signing on the client side (`NineChronicles.Headless.Executable.exe` as CLI itself or [sphere](https://github.com/planetarium/sphere) can help) + `stageTransaction` mutation is more recommendable.
   - Of course, such a combination requires complex scripts and may put users of such scripts in a dangerous environment. However, I believe that the user's private key is more secure than traversing the network.

2. Currently, Crystals are infinite goods that can be produced indefinitely depending on the user's account (to be precise, by the APs produced there), and their transfer is not taken into account at all in the game economy. In terms of game economy, the current situation where crystal transfer is possible is a bug. we expected to solve this problem by resolving the infinite supply of APs, which is the root of the problem, but it seems that it will take time due to complex reasons. So, at least until the end of the year, there is a plan to limit Crystal transferring itself. cc @jaeho0103

PR #1682, opened 2022-11-10
Add ObserveOn on ActionEvaluationPublisher

riemannulus:
Fixed renderers to work in parallel if they were intended to run in parallel but didn't.

### DIFF
--- a/NineChronicles.Headless/ActionEvaluationPublisher.cs
+++ b/NineChronicles.Headless/ActionEvaluationPublisher.cs
@@ -1,3 +1,4 @@
+#nullable enable
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
@@ -8,7 +9,6 @@ using System.Linq;
 using System.Reactive.Concurrency;
 using System.Reactive.Linq;
 using System.Runtime.Serialization;
-using System.Runtime.Serialization.Formatters.Binary;
 using System.Threading;
 using System.Threading.Tasks;
 using Bencodex;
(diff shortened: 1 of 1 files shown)


longfin (approved):

NineChronicles.Headless/ActionEvaluationPublisher.cs
```diff
@@ -103,7 +103,10 @@ public async Task AddClient(Address clientAddress)
             if (!_blockRenderDictionary.ContainsKey(clientAddress))
             {
                 _blockRenderDictionary[clientAddress] = true;
-                _blockRenderer.BlockSubject.SubscribeOn(NewThreadScheduler.Default).Subscribe(
+                _blockRenderer.BlockSubject
+                    .SubscribeOn(NewThreadScheduler.Default)
+                    .ObserveOn(NewThreadScheduler.Default)
```
longfin, line 108: We may need other schedulers (e.g., [TaskPoolScheduler](https://learn.microsoft.com/en-us/previous-versions/dotnet/reactive-extensions/hh229933(v=vs.103)) than `NewThreadSheduler`. 🤔

moreal (approved):

longfin (changes requested):
I'll prepare more fixes because it may cause memory leakage.

moreal (approved):
It seems good to me but I think it'll be better to merge after receiving @ipdae's approval.

NineChronicles.Headless/ActionEvaluationPublisher.cs
```diff
@@ -84,59 +78,169 @@ public async Task AddClient(Address clientAddress)
...
+        {
+            var address = new Address(addressBytes);
+            if (address == default)
+            {
+                Log.Warning("[{ClientAddress}] UpdateSubscribeAddresses set default address", address);
+            }
+            var addresses = addressesBytes.Select(a => new Address(a)).ToImmutableHashSet();
+            if (_clients.TryGetValue(address, out Client? client) && client is { })
+            {
+                client.TargetAddresses = addresses;
```
ipdae, line 126: is this ok not use `TryUpdate`?

longfin: Basically, all `client` in `_clients` are ref types so it will affect objects in the collections without explicit updating. but... yeah, we may need to consider sync behavior.

longfin: I've added https://github.com/planetarium/NineChronicles.Headless/pull/1682/commits/e42b60053092abce9a239241f8ea237785a7b640 to address it. 

Also, I turned this PR into a draft until testing. cc @ipdae @riemannulus @moreal

riemannulus: Confirm.

longfin: Amended. https://github.com/planetarium/NineChronicles.Headless/pull/1682/commits/cf210a1a3486cb666368b6a333b32c2e2044686e cc @riemannulus @moreal @ipdae

ipdae (approved):

Issue #1695, opened 2022-11-16
Subscriptions on `StandaloneContext` make render slower.


longfin:
Related: #1694

PR #1700, opened 2022-11-16
Release v100321-3

longfin:
This PR is a hotfix to address delayed block sync due to #1695 and #1694 .  it isn't a direct fix but it loosens a delay by GQL subscriptions via the two below approaches.

1. Specifies rendering thread for subscriptions in `StandaloneSubscription`, using `ObserveOn()`.
2. In `StandaloneSubscription.RenderBlock`, it introduces `AsParallel()` and `ForAll()` to enable parallelism.

Also, it adds a GQL query to debug this symptom and maintain versions.


ipdae (approved):
As far as I know, all MonsterCollect related actions have been replaced with Stake actions. we can delete renderers after investigation.

PR #1708, opened 2022-11-18
Add timestamp to actionTxQuery

longfin:
This PR adds `timestamp` argument to `actionTxQuery` GraphQL query to specify `Transaction<T>.Timestamp`.


moreal (approved):
🚀

moreal:
It seems there is lint problem. Could you run `dotnet format --exclude Lib9c --exclude NineChronicles.RPC.Shared -v=d --no-restore` once?

longfin:
@moreal Thanks a lot 😄

PR #1712, opened 2022-11-21
Introduce "runeEnhancement" graphql endpoint

U-lis:
To run and test `runeEnhancement` action, a graphql endpoint is created.

This action uses Action factory that selects right version of action by current block index.

- Help Wanted: We need to test this action but I don't know how to test factory based action.

[PR on lib9c](https://github.com/planetarium/lib9c/pull/1517) needs to be merged first

### DIFF
--- a/NineChronicles.Headless/GraphTypes/ActionQuery.cs
+++ b/NineChronicles.Headless/GraphTypes/ActionQuery.cs
@@ -10,9 +10,11 @@ using Libplanet;
 using Libplanet.Assets;
 using Libplanet.Explorer.GraphTypes;
 using Nekoyume.Action;
+using Nekoyume.Action.Factory;
 using Nekoyume.Helper;
 using Nekoyume.Model.State;
 using Nekoyume.TableData;
+using Log = Serilog.Log;
 using NCAction = Libplanet.Action.PolymorphicAction<Nekoyume.Action.ActionBase>;
 
 namespace NineChronicles.Headless.GraphTypes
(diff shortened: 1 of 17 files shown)


NineChronicles.Headless/GraphTypes/ActionQuery.cs
```diff
@@ -347,6 +349,56 @@ public ActionQuery(StandaloneContext standaloneContext)
...
+                        return null;
+                    }
+
+                    var swarm = service.Swarm;
+                    if (!(swarm?.BlockChain is { } chain))
+                    {
+                        throw new InvalidOperationException($"{nameof(swarm.BlockChain)} is null.");
+                    }
+
+                    var tip = swarm.BlockChain.Tip.Index;
```
moreal, line 387:
```suggestion
                    var tip = chain.Tip.Index;
```

NineChronicles.Headless/GraphTypes/ActionQuery.cs
```diff
@@ -347,6 +349,56 @@ public ActionQuery(StandaloneContext standaloneContext)
...
+                    }
+                ),
+                resolve: context =>
+                {
+                    var service = standaloneContext.NineChroniclesNodeService;
+                    if (service is null)
+                    {
+                        Log.Error($"{nameof(NineChroniclesNodeService)} is null.");
+                        return null;
+                    }
```
moreal, line 379: Because the type of `runeEnhancement` is a *non-null* type, it will cause errors. How about throwing errors instead?

NineChronicles.Headless/GraphTypes/States/Models/Table/MonsterCollectionSheetType.cs
```diff
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
```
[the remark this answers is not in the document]
U-lis: What the.. I surely thought deleted this :confused:  Thanks for pointing it!

U-lis:
This will be newly merged to `2022q4-rune` branch soon.
Close without merge.

PR #1717, opened 2022-11-23
Implement ClaimStakeReward interface and factory

U-lis:
NIneChronicles.Headless now implements factory and interface of ClaimStakeReward.

Also, few updates to cli commands.

### DIFF
--- a/NineChronicles.Headless.Executable/Commands/TxCommand.cs
+++ b/NineChronicles.Headless.Executable/Commands/TxCommand.cs
@@ -8,11 +8,13 @@ using Bencodex.Types;
 using Cocona;
 using CsvHelper;
 using Libplanet;
+using Libplanet.Action;
 using Libplanet.Assets;
 using Libplanet.Blocks;
 using Libplanet.Crypto;
 using Libplanet.Tx;
 using Nekoyume.Action;
+using Nekoyume.Action.Factory;
 using Nekoyume.Model;
 using Nekoyume.Model.State;
 using Nekoyume.TableData;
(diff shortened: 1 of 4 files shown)


NineChronicles.Headless.Executable/Commands/TxCommand.cs
```diff
@@ -65,10 +70,11 @@ public TxCommand(IConsole console)
...
                     nameof(ClaimMonsterCollectionReward) => new ClaimMonsterCollectionReward(),
                     nameof(Stake) => new Stake(),
-                    // FIXME: This `ClaimStakeReward` cases need to reduce to one case.
-                    nameof(ClaimStakeReward1) => new ClaimStakeReward1(),
-                    nameof(ClaimStakeReward) => new ClaimStakeReward(),
-                    nameof(ClaimStakeReward3) => new ClaimStakeReward3(),
+                    // FIXME: How to set default action version? Can I use blockIndex instead of action version?
+                    "ClaimStakeReward" => (ActionBase)ClaimStakeRewardFactory.CreateByVersion(
+                        actionVersion ?? 2,
+                        new Address()
```
moreal, line 76: Is it okay to pass an empty address, `new Address()`? I'm not sure why it needs an address to instantiate actions. 🤔 

(It may be overwritten by `LoadPlainValue` 🤔)

moreal (approved):

NineChronicles.Headless.Executable/Commands/TxCommand.cs
```diff
@@ -65,10 +70,11 @@ public TxCommand(IConsole console)
...
+                    ),
```
boscohyun, line 77: I think a default version is not exist. So.. how about this?

```suggestion
                    "ClaimStakeReward" => actionVersion
                        ? (ActionBase)ClaimStakeRewardFactory.CreateByVersion(
                            actionVersion,
                            new Address())
                        : new ClaimStakeReward(new Addess()),
```

There is a comment in `ClaimStakeRewardFactory.CreateByBlockIndex()` like below. And it make some confuse. I'll fix this!

```cs
// FIXME: This method should consider the starting block index of
//        `claim_stake_reward2`. And if the `blockIndex` is less than
//        the starting block index, it should throw an exception.
// default: Version 2
return new ClaimStakeReward(avatarAddress);
```

NineChronicles.Headless.Executable/Commands/TxCommand.cs
boscohyun: @moreal Make sense. Let's fix this from the factory method later!

U-lis: I thought the random address will be overwritten from given plainvalue

PR #1725, opened 2022-11-25
Introduce runeEnhancement GQL endpoint

U-lis:
SSIA


moreal (approved):
I don't know the rune part well but it seems good in the GraphQL part.

PR #1729, opened 2022-11-28
Add `CurrencyEnumType.RUNESTONE`

boscohyun:
- Add `CurrencyEnumType.RUNESTONE`

### DIFF
--- a/NineChronicles.Headless.Tests/GraphTypes/ActionQueryTest.cs
+++ b/NineChronicles.Headless.Tests/GraphTypes/ActionQueryTest.cs
@@ -243,11 +243,16 @@ namespace NineChronicles.Headless.Tests.GraphTypes
         }
 
         [Theory]
-        [InlineData("NCG", true)]
-        [InlineData("NCG", false)]
-        [InlineData("CRYSTAL", true)]
-        [InlineData("CRYSTAL", false)]
-        public async Task TransferAsset(string currencyType, bool memo)
+        [InlineData("NCG", true, true)]
+        [InlineData("NCG", false, true)]
+        [InlineData("CRYSTAL", true, true)]
+        [InlineData("CRYSTAL", false, true)]
+        [InlineData("RUNESTONE", true, false)]
+        [InlineData("RUNESTONE", false, false)]
+        public async Task TransferAsset(
+            string currencyType,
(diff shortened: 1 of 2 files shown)


NineChronicles.Headless/GraphTypes/ActionQuery.cs
```diff
@@ -182,6 +182,7 @@ public ActionQuery(StandaloneContext standaloneContext)
                             (Dictionary)standaloneContext.BlockChain!.GetState(GoldCurrencyState.Address)
                         ).Currency,
                         CurrencyEnum.CRYSTAL => CrystalCalculator.CRYSTAL,
+                        CurrencyEnum.RUNESTONE => throw new ExecutionError("Unsupported Currency type."),
```
moreal, line 185: It's just a question. If it is not supported, is there a reason to add the `CurrencyEnum.RUNESTONE` now?

boscohyun: @moreal I agree. I'll fix it.

moreal: I still don't know what this pull request achieves. What is the use case of the `CurrencyEnum.RUNESTONE` value? 🤔

boscohyun: I was thinking of using it in graphql dealing with currency. It is true that there is no place to use it now. But won't it be possible to use it in the future? Shouldn't `CurrencyEnumType' contain all the currencies defined in lib9c, even if they are not in use right now?
I don't have a strong opinion either, but if I had to define what this change was supposed to accomplish, the questions above would need answers.
If we limit the role of `CurrencyEnumType` to "the query type used", then this pull request should be closed. If we set it to "query type corresponding to the currency type defined in lib9c", we will have to reflect lib9c changes every time.
Right now, I don't think it matters.

boscohyun: On the other hand, reflecting lib9c's currency into `CurrencyEnumType` every time doesn't seem to be smooth every time. This is because the number of points to manage increases when the design of the game currency (e.g., ticker) changes.
For that reason, I'm closing this pull request.

PR #1737, opened 2022-11-29
Provide `appsettings.development.json`

moreal:
Run `dotnet run -- --config appsettings.development.json`

### DIFF
--- a/NineChronicles.Headless.Executable/NineChronicles.Headless.Executable.csproj
+++ b/NineChronicles.Headless.Executable/NineChronicles.Headless.Executable.csproj
@@ -39,6 +39,9 @@
     <None Update="appsettings.json">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Update="appsettings.development.json">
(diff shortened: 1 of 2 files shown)


U-lis (commented):
According to my experience, if I run in as-is state, the dotnet will raise error like this:
```
Unhandled Exception: System.IO.FileNotFoundException: The configuration file 'appsettings.local.json' was not found and is not optional.
```
To resolve this error, we have to add an item to `NineChronicles.Headless.Executable.csproj` file.
But I think this is kinda local setting so I'm not sure this is ok to add all items to csproj file. :thinking:

moreal:
> To resolve this error, we have to add an item to `NineChronicles.Headless.Executable.csproj` file.
> But I think this is kinda local setting so I'm not sure this is ok to add all items to csproj file. 🤔

I understood that as it will occur an error because the build output will not include the `appsettings.development.json`. I think it can be excluded from the build output because it is a kinda local setting as you said.

moreal:
But it'll be useful when running the headless in the published Docker image. I'll apply it.

U-lis (approved):

PR #1761, opened 2022-12-08
[This PR is for samples only]Introduce changeAvatarName action query

boscohyun:
Introduce changeAvatarName action query.


moreal:
This pull request seems stale. 👀

PR #1777, opened 2022-12-19
Unlink Lib9c.Tools to prepare deprecation

moreal:
Since #1602, `Lib9c.Tools` commands were moved from the `lib9c` repository to the `NineChronicles.Headless.Exectuable` command project. And `Lib9c.Tools` became deprecated. But the `NineChronicles.Headless` project still has references to `Lib9c.Tools`. This pull request removes them.


U-lis (approved):
Thanks for all your efforts!

PR #1807, opened 2023-01-04
Ci/build trigger

sky1045:
;;;;

### DIFF
--- a/.github/workflows/push_docker_image.yml
+++ b/.github/workflows/push_docker_image.yml
@@ -62,7 +62,7 @@ jobs:
           docker manifest push $DOCKER_REPO:git-${{ github.sha }}
 
   tag:
-    if: github.ref_type == 'tag' || github.events.inputs.imageTag != ''
+    if: ${{ github.ref_type == 'tag' || github.events.inputs.imageTag != '' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/push_docker_image.yml
```diff
@@ -73,8 +73,8 @@ jobs:
       - name: push git tagged version
         run: |
           docker pull $DOCKER_REPO:git-${{ github.sha }}-amd64
-          if [[ -z ${{ github.events.inputs.imageTag }} ]]; then
-            export IMAGE_TAG=${{ github.events.inputs.imageTag }}
+          if [[ -z ${{ github.event.inputs.imageTag }} ]]; then
+            export IMAGE_TAG=${{ github.event.inputs.imageTag }}
           else
```
moreal, line 78: Does it set `IMAGE_TAG` when the `imageTag == ''`?

```
$ if [[ -z "" ]]; then echo 1; fi 
1
$ if [[ -z "sometag" ]]; then echo 1; fi
# no output
```

sky1045: oooh, i confused, thanks

moreal (approved):

PR #1808, opened 2023-01-04
Minimize Sentry SampleRate only for errors

sky1045:


moreal (commented):
If `SentryTraceSampleRate` will be not used, how about removing the option and related functions?

NineChronicles.Headless.Executable/Program.cs
```diff
@@ -242,9 +242,7 @@ static void ConfigureSentryOptions(SentryOptions o)
                 //o.Debug = true;
                 o.Release = Assembly.GetExecutingAssembly().GetCustomAttribute<AssemblyInformationalVersionAttribute>()
                     ?.InformationalVersion ?? "Unknown";
-                o.SampleRate = headlessConfig.SentryTraceSampleRate > 0
-                    ? (float)headlessConfig.SentryTraceSampleRate
-                    : 0.01f;
+                o.SampleRate = 0.01f;
                 o.TracesSampleRate = headlessConfig.SentryTraceSampleRate;
```
sky1045, line 246: @moreal that option is still used, but only for trace

moreal: Ah, it is my mistake. Thank you 🙏🏻

moreal (approved):

PR #1835, opened 2023-01-18
Improve solution and project settings for dev extensions

boscohyun:
- Related with https://github.com/planetarium/lib9c/pull/1665

This pull request improve NineChronicles.Headless.Executable solution and projects.

- Add `Debug DevEx` configuration based on `Debug`.
- Use `LIB9C_DEV_EXTENSIONS` define.


.editorconfig
```diff
@@ -33,3 +33,6 @@ dotnet_diagnostic.SA1601.severity = suggestion
 
 # Allow FIXME tag as position free
 dotnet_diagnostic.S1134.severity = none
+
+[*.{csproj,xml}]
+indent_size = 2
```
moreal, line 38: Though it configures `indent_size` as `2` but it seems there are `csproj`s which don't follow the rule. It'll be better to add a new GH Actions workflow to check editorconfig.

moreal (approved):

.editorconfig
boscohyun: @moreal That's cool!
https://github.com/planetarium/lib9c/issues/1674

PR #1837, opened 2023-01-18
Add graphql endpoints to essential game actions

U-lis:
- HackAndSlash
- HackAndSlashSweep
- DailyReward
- CombinationEquipment
- CombinationContumable
- RapidCombination
- ItemEnhancement

Those GQL fields are seperated to partial class for readability.


NineChronicles.Headless/GraphTypes/ActionQueryFields/HackAndSlash.cs
```diff
@@ -0,0 +1,102 @@
...
+                        Equipments = equipmentIds,
+                        Foods = consumableIds,
+                        RuneInfos = runeSlotInfos,
+                        StageBuffId = stageBuffId
+                    };
+                    return Encode(context, action);
+                }
+                catch (Exception e)
+                {
+                    var msg = $"Unexpected exception occurred during {typeof(ActionMutation)}: {e}";
```
boscohyun, line 95: Could you tell me why the exception message contains the type of `ActionMutation`?

boscohyun (approved):

NineChronicles.Headless/GraphTypes/ActionQuery.cs
```diff
@@ -19,7 +19,7 @@
 
 namespace NineChronicles.Headless.GraphTypes
 {
-    public class ActionQuery : ObjectGraphType
+    public partial class ActionQuery : ObjectGraphType
```
longfin, line 22: Frankly, I'm not a big fan of `partial` class since usually it can be a signal of abstraction lacking. (i.e., if we need to `partial` to maintain some class, It can also mean that the class has too many features)

it might be better to consider subgraphs, even if on the next PR. 🤔

NineChronicles.Headless/GraphTypes/ActionQueryFields/CombinationConsumable.cs
```diff
@@ -0,0 +1,48 @@
...
+            resolve: context =>
+            {
+                var avatarAddress = context.GetArgument<Address>("avatarAddress");
+                var slotIndex = context.GetArgument<int>("slotIndex");
+                var recipeId = context.GetArgument<int>("recipeId");
+                NCAction action = new CombinationConsumable
+                {
+                    avatarAddress = avatarAddress,
+                    slotIndex = slotIndex,
+                    recipeId = recipeId
```
longfin, line 42:
```suggestion
                    recipeId = recipeId,
```

additionally, it also would be nice with linter to check [SA1413](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/2f6d02f95d34a771717d9d4dab5d1b8819e3e5dd/documentation/SA1413.md).

NineChronicles.Headless/GraphTypes/ActionQueryFields/HackAndSlashSweep.cs
```diff
@@ -0,0 +1,97 @@
...
+                {
+                    Address avatarAddress = context.GetArgument<Address>("avatarAddress");
+                    int worldId = context.GetArgument<int>("worldId");
+                    int stageId = context.GetArgument<int>("stageId");
+                    List<Guid> costumeIds = context.GetArgument<List<Guid>?>("costumeIds") ?? new List<Guid>();
+                    List<Guid> equipmentIds = context.GetArgument<List<Guid>?>("equipmentIds") ?? new List<Guid>();
+                    List<RuneSlotInfo> runeSlotInfos =
+                        context.GetArgument<List<RuneSlotInfo>?>("runeSlotInfos") ?? new List<RuneSlotInfo>();
+                    int actionPoint = context.GetArgument<int>("actionPoint");
+                    int? apStoneCount = context.GetArgument<int?>("apStoneCount");
```
longfin, line 73:
```suggestion
                    int apStoneCount = context.GetArgument<int?>("apStoneCount") ?? 0;
```

Trivial fix for consistency 😅

longfin (commented):
I left a few comments. minor things are fine, but there are things that seem intractable right away in this PR, so feel free to let me know.

NineChronicles.Headless/GraphTypes/ActionQuery.cs
U-lis: Just as you said, there are too many features to put altogether into the one class.
But I wonder subgraph is the answer cuz all actionQueries are on the same hierarchy level I think.
I'll make an issue based on this command and we can continue discussion

NineChronicles.Headless/GraphTypes/ActionQueryFields/HackAndSlash.cs
U-lis: Oh, I just copied that `catch` block from `ActionMutation.cs` and forgot to change name to right one
Thanks for pointing out!

PR #1858, opened 2023-02-01
add rate limit feature

area363:
This PR incorporates a rate limit capability into the GQL server, with a default setting of 10 requests per second per IP.

The rate limit can be enabled via the --rate-limit option, and the configuration can be adjusted through the use of environment variables (e.g. IpRateLimiting__HttpStatusCode=500). By default, this feature is turned off.

### DIFF
--- a/NineChronicles.Headless.Executable/Program.cs
+++ b/NineChronicles.Headless.Executable/Program.cs
@@ -308,7 +308,7 @@ namespace NineChronicles.Headless.Executable
                         HttpOptions = headlessConfig.RpcServer && headlessConfig.RpcHttpServer == true
                             ? new GraphQLNodeServiceProperties.MagicOnionHttpOptions(
                                 $"{headlessConfig.RpcListenHost}:{headlessConfig.RpcListenPort}")
-                            : (GraphQLNodeServiceProperties.MagicOnionHttpOptions?)null
+                            : (GraphQLNodeServiceProperties.MagicOnionHttpOptions?)null,
                     };
 
                     var graphQLService = new GraphQLService(graphQLNodeServiceProperties);
--- a/NineChronicles.Headless.Executable/Program.cs
+++ b/NineChronicles.Headless.Executable/Program.cs
@@ -188,9 +188,7 @@ namespace NineChronicles.Headless.Executable
             string? sentryDsn = "",
             [Option(Description = "Trace sample rate for sentry")]
             double? sentryTraceSampleRate = null,
-            [Ignore] CancellationToken? cancellationToken = null,
-            [Option(Description = "Enable rate limiting to control rate of requests.")]
-            bool? rateLimit = null
+            [Ignore] CancellationToken? cancellationToken = null
         )
         {
 #if SENTRY || ! DEBUG
@@ -311,7 +309,7 @@ namespace NineChronicles.Headless.Executable
                             ? new GraphQLNodeServiceProperties.MagicOnionHttpOptions(
                                 $"{headlessConfig.RpcListenHost}:{headlessConfig.RpcListenPort}")
                             : (GraphQLNodeServiceProperties.MagicOnionHttpOptions?)null,
-                        IpRateLimitOptions = rateLimit == true ? configuration.GetSection("IpRateLimiting") : null,
+                        IpRateLimitOptions = configuration.GetSection("IpRateLimiting"),
                     };
 
                     var graphQLService = new GraphQLService(graphQLNodeServiceProperties);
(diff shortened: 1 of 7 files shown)


NineChronicles.Headless.Executable/appsettings.json
```diff
@@ -110,5 +110,23 @@
...
+                "Endpoint": "*:/graphql",
+                "Period": "1s",
+                "Limit": 10
+            }
+        ],
+        "QuotaExceededResponse": {
+            "Content": "{{ \"message\": \"Whoa! Calm down, cowboy!\", \"details\": \"Quota exceeded. Maximum allowed: {0} per {1}. Please try again in {2} second(s).\" }}",
+            "ContentType": "application/json",
+            "StatusCode": 429
+        }
```
U-lis, line 130: [Just an opinion] In my understand, `IpRateLimiting` setting is one of headless's options, right? So I suggest moving this settings into `Headless` section.
Welcome any ideas :raised_hands:

U-lis (approved):

NineChronicles.Headless.Executable/appsettings.json
area363: Thanks for you feedback!

The `IPRateLimiting` setting in `appsettings.json` is used by the [AspNetCoreRateLimit](https://github.com/stefanprodan/AspNetCoreRateLimit) package. The setting is retreived here (https://github.com/area363/NineChronicles.Standalone/blob/feat/rate-limit/NineChronicles.Headless.Executable/Program.cs#L314) and passed down to `GraphQLService.cs`.

So, the `IPRateLimiting` setting isn't necessarily a headless option. However, because there should be a way to enable or disable `IPRateLimiting`, I've added the `--rate-limit` to the headless option which is `null` in default (meaning rate limiting is disable).

Let me add `RateLimit` inside the appsettings headless option and request another feedback from you. 🙏

NineChronicles.Headless.Executable/Program.cs
```diff
@@ -188,7 +188,9 @@ static void ConfigureSentryOptions(SentryOptions o)
             string? sentryDsn = "",
             [Option(Description = "Trace sample rate for sentry")]
             double? sentryTraceSampleRate = null,
-            [Ignore] CancellationToken? cancellationToken = null
+            [Ignore] CancellationToken? cancellationToken = null,
+            [Option(Description = "Enable rate limiting to control rate of requests.")]
+            bool? rateLimit = null
```
moreal, line 193: Though it isn't a rule, how about ordering the `Ignore` parameter to the last?
```suggestion
            [Option(Description = "Enable rate limiting to control rate of requests.")]
            bool? rateLimit = null,
            [Ignore] CancellationToken? cancellationToken = null
```

NineChronicles.Headless.Executable/Program.cs
```diff
@@ -307,6 +311,7 @@ static void ConfigureSentryOptions(SentryOptions o)
                             ? new GraphQLNodeServiceProperties.MagicOnionHttpOptions(
                                 $"{headlessConfig.RpcListenHost}:{headlessConfig.RpcListenPort}")
                             : (GraphQLNodeServiceProperties.MagicOnionHttpOptions?)null,
+                        IpRateLimitOptions = rateLimit == true ? configuration.GetSection("IpRateLimiting") : null,
```
moreal, line 314: Because `Configuration.GetSection` returns `null` when the section doesn't exist, who about trying `IpRateLimiting` section without `--rate-limit` flag?

https://learn.microsoft.com/ko-kr/dotnet/api/system.configuration.configuration.getsection?view=dotnet-plat-ext-6.0

NineChronicles.Headless/GraphQLService.cs
```diff
@@ -33,7 +34,7 @@ public class GraphQLService
 
         public const string MagicOnionTargetKey = "magicOnionTarget";
 
-        private GraphQLNodeServiceProperties GraphQlNodeServiceProperties { get; }
+        private static GraphQLNodeServiceProperties? GraphQlNodeServiceProperties { get; set; }
```
moreal, line 37: Since it became a `static` property, `GraphQLService` instances will share unintended `GraphQlNodeServiceProperties`. And it seems there is no part to use `setter`. Shouldn't this part be reverted? 🤔

moreal (approved):

area363:
Thanks for the review! I'm currently doing a final test on the mainnet. If it's good to go, I'll go ahead and merge 🙏

area363:
Finalized this PR. I'll get one more review next week and then merge.

area363:
I'll merge this because this PR was tested in `v100360` and is working as expected.

PR #1873, opened 2023-02-02
fix syntax error

area363:
Add quotes around `${{ github.event.inputs.imageTag }}` due to this error: https://github.com/planetarium/NineChronicles.Headless/actions/runs/4063155194/jobs/7010608950#step:3:72


moreal (approved):
I forgot it, thank you 🙏

area363:
Thanks I'll rewrite the tag (v100360-1) after this merge 👍

moreal:
- [ ] @area363  Ah I ran the workflow manually so there is already v100360-1 docker images. I forgot to check the target branch 😞. But if it will not have conflict between previous image and new image, I think that's okay.

PR #1879, opened 2023-02-05
Avatar subscription

TarogStar:
Ability to subscribe to actions that change action points for an agent's avatars, allowing to see the action points available and changes in experience, level and inventory.  This will allow for less checking for this information, and instead supply it only when needed.


moreal (commented):
Could you stack by cherry-picking only your changes on the branch based on the `development` branch?

PR #1891, opened 2023-02-15
⬆️ Bump lib9c

greymistcube:
⬆️

### DIFF
--- a/Libplanet.Headless/Hosting/LibplanetNodeService.cs
+++ b/Libplanet.Headless/Hosting/LibplanetNodeService.cs
@@ -17,6 +17,7 @@ using Libplanet.Blocks;
 using Libplanet.Crypto;
 using Libplanet.Net;
 using Libplanet.Net.Protocols;
+using Libplanet.Net.Transports;
 using Libplanet.RocksDBStore;
 using Libplanet.Store;
 using Libplanet.Store.Trie;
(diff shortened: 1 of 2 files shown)


moreal (approved):

Libplanet.Headless/Hosting/LibplanetNodeService.cs
```diff
@@ -195,27 +196,31 @@ public class LibplanetNodeService<T> : BackgroundService, IDisposable
...
+                    GetBlockHashesTimeout = TimeSpan.FromSeconds(50),
+                    GetBlocksBaseTimeout = TimeSpan.FromSeconds(5),
+                },
+            };
 
+            var transport = NetMQTransport.Create(
+                Properties.SwarmPrivateKey,
+                appProtocolVersionOptions,
+                hostOptions,
+                swarmOptions.MessageTimestampBuffer).ConfigureAwait(false).GetAwaiter().GetResult();
```
longfin, line 218: We may need to care of this case in future PR 🤔

longfin (approved):

Libplanet.Headless/Hosting/LibplanetNodeService.cs
longfin: See also: https://stackoverflow.com/a/46027058

greymistcube: I agree with the sentiment. Not sure about what the best solution would be. I'm leaning towards sync-over-async factory tho. 🙄

PR #1897, opened 2023-02-23
fix: build errors

boscohyun:
- Bump `Lib9c`.
- Add `Lib9c.Utils` project to solution.


U-lis:
My Rider says `Lib9c.Policy` and `Lib9c.Renderers` projects are also needs to be added to solution to build properly.
Could you check about those projects please? :pray:

boscohyun:
@U-lis Yes, it was. I had seen this error too.🙂
There was an error in `Lib9c` and it was fixed a few minutes ago. And then the error has changed.
- Please see also https://github.com/planetarium/lib9c/pull/1750

U-lis (approved):

PR #1920, opened 2023-03-08
Override polymorphic action's action type loader

moreal:
This pull request requires that https://github.com/planetarium/lib9c/pull/1789 must be merged first.

This pull request makes it override `PolymorphicAction<ActionBase>.ActionTypeLoader`. It uses the `Lib9c.DevExtensions` assembly by the `LIB9C_DEV_EXTENSIONS` definition. Or you can customize with `Headless.ActionTypeLoader.StaticActionTypeLoader` in `appsettings.json`. (See `appsettings-schema.json` for detail).


NineChronicles.Headless.Executable/Program.cs
```diff
@@ -357,8 +357,22 @@ static void ConfigureSentryOptions(SentryOptions o)
 
                 IActionTypeLoader MakeStaticActionTypeLoader() => new StaticActionTypeLoader(
                     Assembly.GetEntryAssembly() is { } entryAssembly
+#if LIB9C_DEV_EXTENSIONS
+                        ? new[]
+                        {
+                            typeof(ActionBase).Assembly,
+                            typeof(Lib9c.DevExtensions.Action.CreateOrReplaceAvatar).Assembly,
```
U-lis, line 364: If we add another DevEx actions, do we have to add all individual actions like this?

U-lis (approved):

NineChronicles.Headless.Executable/Program.cs
moreal: No, it adds the `Lib9c.DevExtensions`' assembly, not only an action. If you add another DevEx action to `Lib9c.DevExtensions` project, it will be added automatically. If you create a new project, it should be added like this.

PR #1950, opened 2023-03-22
Fix startIndex as nullable and use tip index as default

boscohyun:
SSIA


moreal (approved):

moreal:
It doesn't have to apply this comment in this pull request, but as a memo, the *Cocona* library supports `RangeAttribute` to validate arguments and options. So it can be automatically validated by attaching `RangeAttribute` to the options.

- https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.rangeattribute.-ctor?view=net-7.0#system-componentmodel-dataannotations-rangeattribute-ctor(system-type-system-string-system-string) 
- https://github.com/mayuki/cocona#validation

PR #1995, opened 2023-04-14
Feat/grpc rate limit

area363:
This PR adds a RPC rate limiting feature for tx staging. The added options are:

- `--rpc-rate-limiter`: Enables rate limiting on the RPC server (turned off by default).
- `--rpc-rate-limiter-window`: Sets the interval(in seconds) in the rate limiter (set to 5 seconds by default).
- `--rpc-rate-limiter-permit`: Sets the max number of requests in the rate limiter (set to 1 request by default).

**Note**: `--rpc-rate-limiter-window` and `--rpc-rate-limiter-permit` options require the `--rpc-rate-limiter` option to be enabled.

### DIFF
--- a/NineChronicles.Headless.Executable/appsettings.json
+++ b/NineChronicles.Headless.Executable/appsettings.json
@@ -66,6 +66,11 @@
             }
         ]
     },
+    "Logging": {
+        "LogLevel": {
+            "Microsoft": "None"
+        }
+    },
     "Headless": {
         "AppProtocolVersionString": "",
         "GenesisBlockPath": "",
--- a/NineChronicles.Headless.Executable/appsettings.json
+++ b/NineChronicles.Headless.Executable/appsettings.json
@@ -96,7 +96,7 @@
         "EnableRateLimiting": false
     },
     "IpRateLimiting": {
-        "EnableEndpointRateLimiting": true,
+        "EnableRateLimiting": false,
         "StackBlockedRequests": true,
         "RealIpHeader": "X-Real-IP",
         "HttpStatusCode": 429,
--- a/NineChronicles.Headless.Executable/appsettings.json
+++ b/NineChronicles.Headless.Executable/appsettings.json
@@ -92,6 +92,9 @@
         "Confirmations": 0,
         "ChainTipStaleBehaviorType": "reboot"
     },
+    "RpcRateLimiting": {
+        "EnableRateLimiting": false
+    },
     "IpRateLimiting": {
         "EnableEndpointRateLimiting": true,
         "StackBlockedRequests": true,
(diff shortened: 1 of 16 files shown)


area363:
@moreal Could you check this PR when you have time? 😄

NineChronicles.Headless/Properties/GrpcRateLimitOptions.cs
```diff
@@ -0,0 +1,11 @@
+namespace NineChronicles.Headless.Properties
+{
+    public class GrpcRateLimitOptions
+    {
+        public const string GrpcRateLimit = "GrpcRateLimit";
```
moreal, line 5: Where is this constant used?

NineChronicles.Headless/GraphQLService.cs
```diff
@@ -170,6 +171,12 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
 
                 app.UseRouting();
                 app.UseAuthorization();
+
+                if (((IList)Environment.GetCommandLineArgs()).Contains("--rpc-rate-limiter"))
+                {
+                    app.UseRateLimiter();
+                }
```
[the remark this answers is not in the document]
area363: Sure, I'll update it so that it can take the `appsettings.json` value as well. 👍

NineChronicles.Headless/Properties/GrpcRateLimitOptions.cs
area363: It's actually not used so I'll delete it. 👍

area363:
@moreal I've applied your suggestions. Thank you for the feedback! 🙇

NineChronicles.Headless.Executable/appsettings.json
```diff
@@ -87,8 +92,11 @@
         "Confirmations": 0,
         "ChainTipStaleBehaviorType": "reboot"
     },
+    "RpcRateLimiting": {
+        "EnableRateLimiting": false
+    },
     "IpRateLimiting": {
-        "EnableEndpointRateLimiting": true,
+        "EnableRateLimiting": false,
```
moreal, line 99: Is it okay to modify too?

https://github.com/stefanprodan/AspNetCoreRateLimit/blob/dc0d24253c8d4762359dae82b559e961a270cd66/src/AspNetCoreRateLimit/Models/RateLimitOptions.cs#L58

area363: https://github.com/planetarium/9c-infra/blob/main/9c-main/chart/templates/remote-headless.yaml#L110
For some reason, it's been working fine with the wrong configuration... 😅 I'll adjust the values to match the package. Thanks!

moreal (approved):

area363:
I'm reverting this PR to update the rate-limiter tested last week and most of this PR is no longer required.

PR #1999, opened 2023-04-23
Combine grpc/gql rate limiter

area363:
A much simpler update of the original rate limiter(https://github.com/planetarium/NineChronicles.Headless/pull/1995) that can be controlled by modifying "appsettings.json".

### DIFF
--- a/NineChronicles.Headless/Middleware/CustomRateLimitMiddleware.cs
+++ b/NineChronicles.Headless/Middleware/CustomRateLimitMiddleware.cs
@@ -1,0 +1,61 @@
+using System.IO;
+using System.Threading.Tasks;
+using AspNetCoreRateLimit;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Options;
+using Serilog;
+using ILogger = Serilog.ILogger;
+
+namespace NineChronicles.Headless.Middleware
+{
+    public class CustomRateLimitMiddleware : RateLimitMiddleware<CustomIpRateLimitProcessor>
+    {
+        private readonly ILogger _logger;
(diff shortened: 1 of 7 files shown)


area363:
@moreal Could you take a look at this PR? It's an update from the previous gRPC rate limit PR(#1995) in order to manage gRPC and gql rate limiter from a single source(`appsettings.json`). FYI, #1995 was no longer required so I reverted it in #1998. Thanks!

NineChronicles.Headless/Middleware/CustomRateLimitMiddleware.cs
```diff
@@ -0,0 +1,61 @@
...
+                httpContext.Request.Body.Seek(0, SeekOrigin.Begin);
+                if (body.Contains("stageTransaction"))
+                {
+                    identity.Path = "/graphql/stagetransaction";
+                }
+
+                return identity;
+            }
+
+            identity.ClientIp = "127.0.0.1";
```
moreal, line 57: Is there a reason to update the client IP to localhost manually?

area363: Oh, this was to manually "whitelist" other http protocols so that they wouldn't be rate limited. But, I think they won't be affected by these rules(https://github.com/planetarium/NineChronicles.Headless/pull/1999/files#diff-e47fc585cf3ec48a24eabba8519cf4fcffbfc15f80cc907dec2379af8ebdf134R102-R120) anyway 😅. I'll take this one out. Thanks for checking!

moreal (approved):

PR #2012, opened 2023-05-02
Implement `RemoteActionEvaluator`

moreal:
The `RemoteActionEvaluator` implements the `Libplanet.Action.IActionEvaluator` interface. It is to assume the situation that Libplanet cannot access Lib9c's DLL directly. extremely.


moreal:
I'm implementing this feature alone, which doesn't affect the previous implementations. So I'll merge when the build passed.

PR #2016, opened 2023-05-08
GQL Query to see Craft Slot Data

FioX0:
A GQL Query that allows a user to pull data from the crafting slot.

This would allow users to collect the index of when crafting/upgrading is done. Thereby users that use GQL won't have to send failed craft/upgrading requests to verify if the slot is empty. Simply check the slot with this query and you will know.

`{
  "data": {
    "stateQuery": {
      "combinationSlot": {
        "address": "0x5348540744bA52d6012CDA62566fa8746d877F38",
        "unlockBlockIndex": 6758172
      }
    }
  },
  "extensions": {}
}`

### DIFF
--- a/NineChronicles.Headless.Tests/GraphTypes/ActionQueryTest.cs
+++ b/NineChronicles.Headless.Tests/GraphTypes/ActionQueryTest.cs
@@ -878,15 +878,21 @@ actionPoint: {actionPoint},
         }
 
         [Fact]
-        public async Task ItemEnhantement()
+        public async Task ItemEnhancement()
         {
             var avatarAddress = new PrivateKey().ToAddress();
             var slotIndex = 0;
             var itemId = Guid.NewGuid();
-            var materialId = Guid.NewGuid();
+            var materialIds = new List<Guid> { Guid.NewGuid() };
 
+            var materialQuery = new StringBuilder("[");
+            foreach (var materialId in materialIds)
+            {
+                materialQuery.Append($" \"{materialId}\"");
+            }
+            materialQuery.Append("]");
             var query = $"{{itemEnhancement(avatarAddress: \"{avatarAddress}\", slotIndex: {slotIndex}, " +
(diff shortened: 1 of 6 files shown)


NineChronicles.Headless/GraphTypes/StateQuery.cs
```diff
@@ -514,7 +514,35 @@ public StateQuery()
...
+                    new QueryArgument<NonNullGraphType<IntGraphType>>
+                    {
+                        Name = "slot",
+                        Description = "Slot index 0-3"
+                    }
+                ),
+                resolve: context =>
+                {
+                    var agentAddress = context.GetArgument<Address>("avatarAddress");
+                    var index = context.GetArgument<int>("slot");
```
longfin, line 536: Since `index` is only valid when between 0 and 3, checking and throwing an error against illegal values might be nice. 😄

longfin (approved):

NineChronicles.Headless/GraphTypes/StateQuery.cs
FioX0: I'll add that 👍

FioX0: added it now, sorry for delay. Had my system drive fail so had to install the whole OS again.

NineChronicles.Headless/GraphTypes/StateQuery.cs
```diff
@@ -514,7 +514,38 @@ public StateQuery()
...
+                    if(index < 0 || index > 3)
+                    {
+                        throw new Exception("Invalid Slot Index");
+                    }
+                    var deriveAddress = CombinationSlotState.DeriveAddress(agentAddress, index);
```
ipdae, line 541:
```suggestion
                    var deriveAddress = CombinationSlotState.DeriveAddress(avatarAddress, index);
```

PR #2070, opened 2023-05-31
Add requiredBlockIndex field

ipdae:
add requiredBlockIndex field.
it can be use like this.
```
query {
  stateQuery {
    avatar(avatarAddress: "") {
      inventory {
        equipments {
          requiredBlockIndex
        }
      }
    }
  }
}
```


longfin (approved):

moreal (approved):
Sometimes.. we should use [union feature](https://graphql-dotnet.github.io/docs/getting-started/unions/). (But it may be a breaking change 😢 )

PR #2076, opened 2023-06-12
downgrade headless to `.NET6`

area363:
This PR downgrades the headless target framework to `.NET6` because we've found that the `.NET7` version occasionally results in a segmentation fault(https://github.com/dotnet/runtime/issues/72365) which degrades our live RPC service quality. `.NET6` headless doesn't have this issue.

According to https://github.com/dotnet/runtime/issues/72365, this issue is fixed from `v7.0.7` so I will do another test when `v7.0.7` is officially released(not yet released).

@moreal Would it be ok to remove `Microsoft.AspNetCore.OpenApi` in `Lib9c.StateService` because this nuget package is only available starting from `.NET7`? It build successfully but I'm not sure if removing it would affect `Lib9c.StateService`'s feature. 🤔

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Setup .NET Core
         uses: actions/setup-dotnet@v3
         with:
-          dotnet-version: 7.0.*
+          dotnet-version: 6.0.x
       - name: Install dependencies
         run: dotnet restore
       - name: Build and test
--- a/.github/workflows/deploy_gh_pages.yml
+++ b/.github/workflows/deploy_gh_pages.yml
@@ -18,7 +18,7 @@ jobs:
       - uses: actions/setup-dotnet@v3
         name: Set up .NET Core SDK
         with:
-          dotnet-version: 6.0.*
+          dotnet-version: 6.0.x
       - name: Build GraphQL Schema
         run: |
           dotnet restore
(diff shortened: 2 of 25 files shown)


area363:
@moreal Also, do you know why dotnet format runs on 7.0(https://github.com/area363/NineChronicles.Standalone/actions/runs/5239663011/jobs/9459707384#step:4:13) even when it's set to 6.0(https://github.com/planetarium/NineChronicles.Headless/pull/2076/files#diff-107e910e9f2ebfb9a741fa10b2aa7100cc1fc4f5f3aca2dfe78b905cbd73c0d2R13)?

moreal:
> Also, do you know why dotnet format runs on 7.0(https://github.com/area363/NineChronicles.Standalone/actions/runs/5239663011/jobs/9459707384#step:4:13) even when it's set to 6.0(https://github.com/planetarium/NineChronicles.Headless/pull/2076/files#diff-107e910e9f2ebfb9a741fa10b2aa7100cc1fc4f5f3aca2dfe78b905cbd73c0d2R13)?

Well... I have to see more. The setup-dotnet works well but the format uses another one. We have to find a way to specify dotnet version in CLI like global.json does.

> Would it be ok to remove `Microsoft.AspNetCore.OpenApi` in `Lib9c.StateService` because this nuget package is only available starting from `.NET7`? It build successfully but I'm not sure if removing it would affect `Lib9c.StateService`'s feature. 🤔

It was used because dotnet-template already has it. I guess it will be okay if the build was successful.

moreal:
@area363 Maybe we can introduce global.json file

https://github.com/planetarium/libplanet/blob/main/global.json

PR #2080, opened 2023-06-15
Automate validating GitHub Actions workflow

moreal:
Though I approved #2079 and it was merged, the workflow was broken. This pull request automates validating GitHub Actions with actionlint. It checks GitHub Actions workflow YAML schema and runs shellcheck for `steps.run` scripts.


area363 (approved):
🙇

PR #2082, opened 2023-06-15
Add IP ban feature for abusers

area363:
This PR bans IPs that send too many txs to the RPC nodes and works concurrently with the original rate-limiter.

Taking the default configuration(https://github.com/planetarium/NineChronicles.Headless/compare/development...area363:feat/add-ip-ban?expand=1#diff-e47fc585cf3ec48a24eabba8519cf4fcffbfc15f80cc907dec2379af8ebdf134R117-R123) as an example, the headless node will block IP's that send `10` more txs than the default rate-limit threshold(`12 transactions` per `60 seconds`) and it will block the IP for `60 minutes` (the IP will be unbanned after `60 minutes`).

### DIFF
--- a/NineChronicles.Headless/Properties/IpBanResponse.cs
+++ b/NineChronicles.Headless/Properties/IpBanResponse.cs
@@ -1,0 +1,12 @@
+namespace NineChronicles.Headless.Properties
+{
+    public class IpBanResponse
+    {
+        public string? ContentType { get; set; }
+
+        public string? Content { get; set; }
+
+        public int? StatusCode { get; set; } = 403;
+    }
+
+}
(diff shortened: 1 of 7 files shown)


area363:
@moreal Could you take a look at this? 🙏

moreal:
> @moreal Could you take a look at this? 🙏

@area363 I'll take a review until 19:00 KST today. Thank you for reminding 🙏🏻

NineChronicles.Headless/Properties/IpBanResponse.cs
```diff
@@ -0,0 +1,12 @@
...
+{
+    public class IpBanResponse
+    {
+        public string? ContentType { get; set; }
+
+        public string? Content { get; set; }
+
+        public int? StatusCode { get; set; } = 403;
+    }
+
```
moreal, line 11: It is not critical comment.

```suggestion
```

moreal (approved):
🚫 👍🏻

NineChronicles.Headless/Properties/IpBanResponse.cs
area363: Fixed! Thank you.

moreal (approved):

PR #2084, opened 2023-06-16
Arena GQL Simulator - Win % calculator

FioX0:
Please note that this might be a bit of a resource intensive query. It performs quite well on my Dev PC, but on my VPS I do see a bit of a slowdown but the VPS isn't really strong to begin with.

I have optimized this as much as I could. Might not be suitable for production, but letting planetarium to make that decision.

This allows you to provide 2 AvatarAddresses and it will simulate 1000 fights and return a decimal which would be the % of avatar1 winning against Avatar2.

`query{
  stateQuery{
    arenaPercentageCalculator(avatarAddress:"0x3b7a47daaece48807fc00a310b05bd9f5d26736e", enemyAvatarAddress:"0xab44635462880666daa7f2be5a21c71c1590ff2b")
  }
}`

`{
  "data": {
    "stateQuery": {
      "arenaPercentageCalculator": 3
    }
  },
  "extensions": {}
}`


NineChronicles.Headless/GraphTypes/StateQuery.cs
```diff
@@ -346,7 +349,136 @@ public StateQuery()
                     return null;
                 }
             );
+            Field<NonNullGraphType<DecimalGraphType>>(
+                name: "arenaPercentageCalculator",
```
longfin, line 353: I guess `stateQuery` might not be the best place for this sort of query because it seems very domain specific(i.e., PVP system) logic instead of simple state lookups. (but on the other hand, it isn't strong due to a lack of alternatives 😅 ) cc @planetarium/nine-chronicles

NineChronicles.Headless/GraphTypes/StateQuery.cs
```diff
@@ -346,7 +349,136 @@ public StateQuery()
...
+                        enemyArenaEquipementList,
+                        enemyArenaCostumeList,
+                        enemyRuneStates);
+
+                    Random rnd  =new Random();          
+
+                    int win = 0;
+                    int loss = 0;
+
+                    for (var i = 0; i < 1000; i++)
```
longfin, line 463: it would be nice if we have an option for this iteration count as configurable.

NineChronicles.Headless/GraphTypes/StateQuery.cs
```diff
@@ -346,7 +349,136 @@ public StateQuery()
...
+                        if(log.Result.ToString() == "Win")
+                        {
+                            win++;
+                        }
+                        else
+                        {
+                            loss++;
+                        }        
+                    }
+                    return Math.Round(((decimal)win / 1000) * 100m, 2);
```
longfin, line 480: Returning aggregated summary makes sense but I also think that more detailed information might be more helpful.

e.g.,

```json
{
    "blockIndex": 1,
    "result": [
        {
            "seed": 1234,
            "win": true
        },
        {
            "seed": 5678,
            "win": false
        },
    ]
}
```

FioX0: Thanks Swen.

So BlockIndex,
Percentage and then a 
Results List per seed/win state?

Then user can decide if they want the result list?

NineChronicles.Headless/GraphTypes/StateQuery.cs
FioX0: May I suggest a new category called "simulatorQuery"? I also got a Stage/PVE one that is almost done. So both queries could be set under this new category.

NineChronicles.Headless/GraphTypes/StateQuery.cs
longfin: introducing a hard cap would be nice! I don't have an appropriate number for that, but both 1k and 5k are looks good to me.

NineChronicles.Headless/GraphTypes/StateQuery.cs
FioX0: @longfin 
Like so?
`{
  "data": {
    "simulationQuery": {
      "arenaPercentageCalculator": {
        "blockIndex": 7133512,
        "result": [
          {
            "seed": 1893608300,
            "win": true
          },
          {
            "seed": 2073532131,
            "win": true
          },
          {
            "seed": 1826085108,
            "win": false
          },
          {
            "seed": 1985649038,
            "win": false
          },
          {
            "seed": 614892139,
            "win": true
          },
          {
            "seed": 2060699251,
            "win": false
          },
          {
            "seed": 784343549,
            "win": true
          },
          {
            "seed": 2109394172,
            "win": true
          },
          {
            "seed": 1339463095,
            "win": true
          },
          {
            "seed": 1834517228,
            "win": false
          }
        ],
        "winPercentage": 60
      }
    }
  },
  "extensions": {}
}`

longfin: @FioX0 right 😄

PR #2093, opened 2023-06-26
Set gasLimit by each action

ipdae:
- Delete gasLimit arguments and set gasLimit by action type


NineChronicles.Headless/GraphTypes/ActionTxQuery.cs
```diff
@@ -28,7 +28,7 @@ internal override byte[] Encode(IResolveFieldContext context, ActionBase action)
             Address signer = publicKey.ToAddress();
             long nonce = context.Parent!.GetArgument<long?>("nonce") ?? blockChain.GetNextTxNonce(signer);
             DateTimeOffset? timestamp = context.Parent!.GetArgument<DateTimeOffset?>("timestamp");
-            long? gasLimit = context.Parent!.GetArgument<long?>("gasLimit");
+            long? gasLimit = action is ITransferAsset or ITransferAssets ? RequestPledge.DefaultRefillMead : 1L;
```
longfin, line 31: I think that we might need to delete `RequestPledge.DefaultRefillMead` field due to the semantic mismatch. (even in the next PR).

longfin (approved):

moreal (approved):

PR #2103, opened 2023-06-29
Validate tx in grpc PutTransaction

ipdae:
Avoid invalid tx staging.

### DIFF
--- a/NineChronicles.Headless/BlockChainService.cs
+++ b/NineChronicles.Headless/BlockChainService.cs
@@ -91,6 +91,10 @@ namespace NineChronicles.Headless
                         _blockChain.StageTransaction(tx);
                         _swarm.BroadcastTxs(new[] { tx });
                     }
+                    else
+                    {
+                        Log.Debug("Skip StageTransaction({TxId}) reason: {Msg}", tx.Id, validationExc.Message);
+                    }
 
                     span.Finish();
                     sentryTrace.StartChild(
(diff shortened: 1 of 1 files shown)


NineChronicles.Headless/BlockChainService.cs
```diff
@@ -83,8 +83,14 @@ public UnaryResult<bool> PutTransaction(byte[] txBytes)
                 {
                     Log.Debug("PutTransaction: (nonce: {nonce}, id: {id})", tx.Nonce, tx.Id);
                     Log.Debug("StagedTransactions: {txIds}", string.Join(", ", _blockChain.GetStagedTransactionIds()));
-                    _blockChain.StageTransaction(tx);
-                    _swarm.BroadcastTxs(new[] { tx });
+#pragma warning disable CS8632
+                    Exception? validationExc = _blockChain.Policy.ValidateNextBlockTx(_blockChain, tx);
```
moreal, line 87: It will be better to leave error logs (e.g., `_logger.Error`) for debugging. It is not required.

moreal (approved):

NineChronicles.Headless/BlockChainService.cs
ipdae: Add message in https://github.com/planetarium/NineChronicles.Headless/pull/2103/commits/38dc13f43684ed12581fac466ef803c9dfa4fa9d

PR #2112, opened 2023-07-03
⬆️ Bump lib9c to 014bff43

greymistcube:
Passes all tests, but I cannot ensure things didn't break (especially with `RemoteActionEvaluator` and `ReplayCommand`). 😣
There will be other releases for libplanet shortly to mitigate the issues somewhat.


riemannulus (commented):
IMO this should be checked on the Backend team.

riemannulus (approved):

PR #2138, opened 2023-07-18
feat: store updated states' addresses at least

moreal:
This pull request suggests storing `TxSuccess.UpdatedStates`' addresses at least. When users want to know *REAL UPDATED ADDRESSES*, there is no way to know it because `Transaction.UpdatedAddresses` is deprecated and not trustable, and `TxSuccess.UpdatedStates` doesn't store anything. Then why we used `ReducedStore`? The `TxSuccess.UpdatedStates` was too big to store in gamers' storage. (like a copy of the `/states` directory). Sadly, no gamers have local storage. In addition, it is not big to store only updated addresses.

If this feature is applied, developers can get real updated addresses easily since the applied block index.

### DIFF
--- a/Libplanet.Headless/ReducedStore.cs
+++ b/Libplanet.Headless/ReducedStore.cs
@@ -112,7 +112,7 @@ namespace Libplanet.Headless
             TxSuccess reducedTxSuccess = new TxSuccess(
                 txSuccess.BlockHash,
                 txSuccess.TxId,
-                updatedStates: ImmutableDictionary<Address, IValue>.Empty,
+                updatedStates: txSuccess.UpdatedStates.ToImmutableDictionary(pair => pair.Key, _ => (IValue)Null.Value),
                 fungibleAssetsDelta: txSuccess.FungibleAssetsDelta,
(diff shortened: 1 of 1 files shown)


Libplanet.Headless/ReducedStore.cs
```diff
@@ -16,7 +16,7 @@ namespace Libplanet.Headless
     /// <list type="bullet">
     /// <item><description><see cref="PutTxExecution(TxSuccess)"/></description></item>
     /// </list>
-    /// </summary>
+    /// </summary>.
```
area363, line 19: Does this `.` mean anything?

moreal: ~~It is `dotnet format`'s act 🙄~~ I'll revert it. Thanks 🙏🏻

moreal: Amended! https://github.com/planetarium/NineChronicles.Headless/compare/abd3880e5025d5c36a24760f6f0f5ba0a96c4cdf..fee345b9d7540483a27a1b9c00843893ad20e6f6

area363 (approved):

PR #2144, opened 2023-07-19
⬆️ Bump lib9c to d1070661 (libplanet 2.5.0)

greymistcube:
@moreal 
This might further break `RemoteActionEvaluator`'s behavior as this no longer allows access to the states returned that's not part of the `Delta`. 😕

@boscohyun 
You might want to take a look at the changed `StateQueryTest.cs` file (I've left some comments). Aside from the mentioned test, as various getters such as `AccountStateGetter` have been removed, many tests were reimplemented using `MockState`. I'd say most related tests have become more comprehensive. Maybe `AvatarStateTypeTest.cs` might also need a closer look since it was one of the less trivial ones.


Libplanet.Extensions.ActionEvaluatorCommonComponents/AccountStateDelta.cs
```diff
@@ -129,25 +120,18 @@ public AccountStateDelta(byte[] bytes)
...
-            TotalSupplyGetter = TotalSupplyGetter,
-        };
-
+        new AccountStateDelta(_states.SetItem(address, state), _fungibles, _totalSupplies, _validatorSet);
     public FungibleAssetValue GetBalance(Address address, Currency currency)
     {
         if (!_fungibles.TryGetValue((address, currency), out BigInteger rawValue))
         {
-            return BalanceGetter(address, currency);
+            throw new NotSupportedException();
```
moreal, line 134: It can be resolved by `_baseState: IAccountState` as others do. 🤔

moreal (approved):
`Libplanet.Action.State.AccountStateDelta` should be public 🤔

Libplanet.Extensions.ActionEvaluatorCommonComponents/AccountStateDelta.cs
moreal: For now, I will merge and edit separately.

PR #2179, opened 2023-08-07
Add avatars query

### LINKED ISSUE #2175: Add new API to lookup many avatars
Like #1650, we need to GQL query to fetch bulk avatar states at once. (see also #1652)

Furthermore, due to the size of avatar states, we may need to limit the address count. 🤔

XxshiftxX:
This PR adds a `avatars` field to query multiple `avatar` by a bunch of address. It resolves #2175.


U-lis:
IMHO, we can leave old `avatar` query as-is and create new `avatars` query.
Because it may used at here and there I think, so it's better to keep backward compatibility even if it seems very redundant.

longfin (approved):

NineChronicles.Headless/GraphTypes/StateQuery.cs
```diff
@@ -40,18 +56,27 @@ public StateQuery()
...
-                    }
-                    catch (InvalidAddressException)
+                    return GetAvatarState(context.Source, address)
+                        ?? throw new InvalidOperationException($"The state {address} doesn't exists");
+                });
+            Field<NonNullGraphType<ListGraphType<AvatarStateType>>>(
+                name: "avatars",
+                description: "Avatar states having some order as addresses",
+                arguments: new QueryArguments(
+                    new QueryArgument<NonNullGraphType<ListGraphType<AddressType>>>
```
moreal, line 66: If you don't want to receive `[null]`, it should be:

```suggestion
                    new QueryArgument<NonNullGraphType<ListGraphType<NonNullGraphType<AddressType>>>>
```

NineChronicles.Headless/GraphTypes/StateQuery.cs
```diff
@@ -40,18 +56,27 @@ public StateQuery()
...
                     {
-                        throw new InvalidOperationException($"The state {address} doesn't exists");
+                        Name = "addresses",
+                        Description = "Addresses of agent to query avatars."
```
moreal, line 69: It seems to be used as `avatar`'s addresses instead of `agent`'s addresses. Could you update the description?
```suggestion
                        Description = "Addresses of avatars to query."
```

moreal (approved):

PR #2180, opened 2023-08-07
track client count by device

area363:
This PR utilizes `GrpcCaptureMiddleware` to track RPC clients by device (PC/mobile/other). 

In order to use the `GrpcCaptureMiddleware`, `ActionEvaluationPublisher` needs to be passed down to the middleware so I've made some adjustments in `Program.cs` when `UseNineChroniclesRPC()` method is called.

### Device metrics in grafana
image

### DIFF
--- a/NineChronicles.Headless/ActionEvaluationPublisher.cs
+++ b/NineChronicles.Headless/ActionEvaluationPublisher.cs
@@ -75,7 +75,7 @@ namespace NineChronicles.Headless
                 description: "Number of RPC clients connected.");
             meter.CreateObservableGauge(
                 "ninechronicles_rpc_clients_count_by_device",
-                () => new []
+                () => new[]
                 {
                     new Measurement<int>(this.GetClientsCountByDevice("mobile"), new[] { new KeyValuePair<string, object?>("device", "mobile") }),
                     new Measurement<int>(this.GetClientsCountByDevice("pc"), new[] { new KeyValuePair<string, object?>("device", "pc") }),
--- a/NineChronicles.Headless/ActionEvaluationPublisher.cs
+++ b/NineChronicles.Headless/ActionEvaluationPublisher.cs
@@ -44,6 +44,7 @@ namespace NineChronicles.Headless
         private readonly NodeStatusRenderer _nodeStatusRenderer;
 
         private readonly ConcurrentDictionary<Address, Client> _clients = new ConcurrentDictionary<Address, Client>();
+        private readonly ConcurrentDictionary<Address, string> _clientsByDevice = new ConcurrentDictionary<Address, string>();
 
         private RpcContext _context;
         private ConcurrentDictionary<string, Sentry.ITransaction> _sentryTraces;
@@ -72,6 +73,18 @@ namespace NineChronicles.Headless
                 "ninechronicles_rpc_clients_count",
                 () => this.GetClients().Count,
                 description: "Number of RPC clients connected.");
+            meter.CreateObservableGauge(
+                "ninechronicles_rpc_clients_mobile_count",
+                () => this.GetClientsCountByDevice("mobile"),
+                description: "Number of mobile RPC clients connected.");
+            meter.CreateObservableGauge(
+                "ninechronicles_rpc_clients_pc_count",
+                () => this.GetClientsCountByDevice("pc"),
+                description: "Number of pc RPC clients connected.");
+            meter.CreateObservableGauge(
+                "ninechronicles_rpc_clients_other_count",
+                () => this.GetClientsCountByDevice("other"),
+                description: "Number of other RPC clients connected.");
 
             ActionEvaluationHub.OnClientDisconnected += RemoveClient;
         }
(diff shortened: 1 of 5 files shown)


NineChronicles.Headless/ActionEvaluationPublisher.cs
```diff
@@ -72,6 +73,18 @@ public class ActionEvaluationPublisher : BackgroundService
...
+                () => this.GetClientsCountByDevice("mobile"),
+                description: "Number of mobile RPC clients connected.");
+            meter.CreateObservableGauge(
+                "ninechronicles_rpc_clients_pc_count",
+                () => this.GetClientsCountByDevice("pc"),
+                description: "Number of pc RPC clients connected.");
+            meter.CreateObservableGauge(
+                "ninechronicles_rpc_clients_other_count",
+                () => this.GetClientsCountByDevice("other"),
+                description: "Number of other RPC clients connected.");
```
moreal, line 87: How about using `tags` (`labels`) feature instead assigning all metrics?


With tags, it can be presented like the below metrics:

```
process_runtime_dotnet_gc_heap_size_bytes{generation="gen0"} 4414816 1691449546325
process_runtime_dotnet_gc_heap_size_bytes{generation="gen1"} 21489360 1691449546325
process_runtime_dotnet_gc_heap_size_bytes{generation="gen2"} 240 1691449546325
process_runtime_dotnet_gc_heap_size_bytes{generation="loh"} 29081808 1691449546325
process_runtime_dotnet_gc_heap_size_bytes{generation="poh"} 179128 1691449546325
```

In this case, we can use [`CreateObservableGauge(string, Func<IEnumerable<Measuarement<T>>>, string?, string?)`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.meter.createobservablegauge?view=net-7.0#system-diagnostics-metrics-meter-createobservablegauge-1(system-string-system-func((system-collections-generic-ienumerable((system-diagnostics-metrics-measurement((-0))))))-system-string-system-string)) now.

It can be like:

```csharp
var meter = new Meter("NineChronicles");
meter.CreateObservableGauge(
    "ninechronicles_rpc_clients_count",
    () => new []
    {
        new Measurement<int>(this.GetClientsCountByDevice("mobile"), new[] { new KeyValuePair<string, object?>("device", "mobile") }),   
        new Measurement<int>(this.GetClientsCountByDevice("pc"), new[] { new KeyValuePair<string, object?>("device", "pc") }),   
        new Measurement<int>(this.GetClientsCountByDevice("other"), new[] { new KeyValuePair<string, object?>("device", "other") }),   
    },
    description: "Number of RPC clients connected.");
```

Or with LINQ:

```csharp
var meter = new Meter("NineChronicles");
meter.CreateObservableGauge(
    "ninechronicles_rpc_clients_count",
    () => _clientsByDevice.GroupBy(pair => pair.Value)
        .Select(group => new Measurement<int>(group.Count(),
            new[] { new KeyValuePair<string, object?>("device", group.Key) })),
    description: "Number of RPC clients connected.");
```

***LINQ version will not return anything as metrics if there are no clients.***

NineChronicles.Headless/GraphTypes/RpcInformationQuery.cs
```diff
@@ -19,6 +20,34 @@ public RpcInformationQuery(ActionEvaluationPublisher publisher)
...
+                        Name = "device"
+                    }
+                ),
+                description: "total count by connected to this node.",
+                resolve: context =>
+                {
+                    string device = context.GetArgument<string>("device");
+                    return publisher.GetClientsCountByDevice(device);
+                });
+            Field<NonNullGraphType<ListGraphType<AddressType>>>(
```
moreal, line 37: If `GetClientsByDevice` cannot a list including `null`, it can guarantee it in type-level.

```suggestion
            Field<NonNullGraphType<ListGraphType<NonNullGraphType<AddressType>>>>(
```

NineChronicles.Headless/ActionEvaluationPublisher.cs
```diff
@@ -112,6 +125,35 @@ public async Task AddClient(Address clientAddress)
             }
         }
 
+        public void AddClientByDevice(Address clientAddress, string device)
+        {
+            if (!_clientsByDevice.ContainsKey(clientAddress))
+            {
+                _clientsByDevice.TryAdd(clientAddress, device);
```
moreal, line 132: It is just a question 🤔 

1. I played on my PC in RPC mode. (Register my address with "pc" string)
1. I didn't turn off my client yet.
1. I played on my phone in RPC mode. (Do not register my address with "mobile" string because there is already my address in the dictionary)

Then it may be still `("<MY_ADDRESS>", "pc")`. Is it right? (not critical, just checking intention 📝)

moreal (commented):
👍🏻 📱

NineChronicles.Headless/ActionEvaluationPublisher.cs
area363: That's an intereseting question!

There are two cases for this:
1) My PC client is connected to `RPC1` and the mobile client also connects to `RPC1` with the same account.
- In this case, the PC client automatically loses connection and the mobile client gains connection to `RPC1` (This only occurs when you use the same account to connect to the same node. I'm not sure how but it seems to be a default gRPC connection behavior). So the client address will be transferred to the `mobile` category.

2) My PC client is connected to `RPC1` and the mobile client also connects to `RPC2` with the same account.
- In this case, `RPC1` will keep the client address in the `PC` category and `RPC2` will have it in the `mobile` category.

NineChronicles.Headless/ActionEvaluationPublisher.cs
area363: I'll make adjustments on this. Thank you for the feedback!

area363:
@moreal PTAL one more time, thanks! 🙇

moreal (approved):

PR #2182, opened 2023-08-08
Add equipped filter into equipments field

### LINKED ISSUE #2176: Add new field  that holding only equipped items, to avatar state
`AvatarState` type has `inventory` field and it has many `equipment`s there. but in some cases, we want to refer only "equipped" items. (e.g., calculate CP, simulate battle)

It would be nice if we also have the equipped items as a field like `inventory`.

XxshiftxX:
Add argument to equipments to filter it is equipped or not. It resolves #2176.


moreal (approved):

ipdae (commented):
It would be nice to apply it to costumes as well.

PR #2183, opened 2023-08-08
Introduce runes query in avatarState GraphQL

U-lis:
To get runes in avatar, I created new `RuneStateType` and include it into `avatarState` query.
`runes` is not inside inventory because it's not in [Inventory](https://github.com/planetarium/lib9c/blob/development/Lib9c/Model/Item/Inventory.cs), but in [avatar](https://github.com/planetarium/lib9c/blob/07fad34fa36e03af8cf317938812eb2667592b85/Lib9c/Model/State/RuneState.cs#L9-L10).

Query looks like:
```graphql
query {
  stateQuery {
    avatar (avatarAddress:"0xDD21DCc1A9d393550A49999363e383c8409Ee1A2") {
      runes {
        runeId
        level
      }
    }
  }
}
```

and the result looks like:
```json
{
  "data": {
    "stateQuery": {
      "avatar": {
        "runes": [
          {
            "runeId": 10001,
            "level": 9
          },
          {
            "runeId": 10002,
            "level": 79
          },
          {
            "runeId": 10003,
            "level": 42
          },
          ...
        ]
      }
    }
  },
  "extensions": {}
}
```

Screenshot from 2023-08-09 01-05-07

### DIFF
--- a/NineChronicles.Headless/GraphTypes/States/AvatarStateType.cs
+++ b/NineChronicles.Headless/GraphTypes/States/AvatarStateType.cs
@@ -1,11 +1,13 @@
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using Bencodex.Types;
 using GraphQL.Types;
-using Libplanet.Action;
 using Libplanet.Explorer.GraphTypes;
 using Libplanet.Action.State;
+using Nekoyume.Action;
 using Nekoyume.Model.State;
+using Nekoyume.TableData;
 using NineChronicles.Headless.GraphTypes.States.Models;
 using NineChronicles.Headless.GraphTypes.States.Models.World;
 using NineChronicles.Headless.GraphTypes.States.Models.Item;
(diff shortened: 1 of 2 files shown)


moreal (approved):
At the GraphQL API level, I approve. But I am not sure about the rune system. When I looked `RuneEnhancement` action, it seems there are currencies too 🤔 I am not sure what the `Rune list of avatar` should show. Only rune states or both rune states and rune currencies... Of course, according to my memory when I played Nine Chronicles, I can imagine what this GraphQL API wants to show.

I want to request @ipdae to review this pull request 🙏🏻 If he says OK, it'll be okay 👀

U-lis:
@moreal You're right. Runes and runestones are FungibleAssetValue at the beginning of design.
My intent of this query is get the rune as an item. Querying rune as a currency, we can use balance query like attachment.
Screenshot from 2023-08-09 09-54-09

Frankly, I'm not sure this is the best way to show the **state** of the rune. But we don't treat rune as a currency in game logic (especially battle or adventure), so I think this is acceptable level.

I also want to request a review to @ipdae and very thankful if he check this things.

ipdae (commented):
As far as I know,
RUNE_STONE: FungibleAssetValue
RUNE: additional equipment to equip in RuneSlot
RuneState: list of owned runes(like inventory)
RuneSlotState: list of equipped runes by battle type(arena, adventure, raid)

If this pr providing about the avatar's arena information, I think RuneSlotStateType is also needed.

U-lis:
RuneSlotState is not very required at the moment to get battle-related data.
Because we need to get rune's state(id, level), not the slot's state(slotType, isLocked).
Someday we may need RuneSlotState from GQL too, so I'll make this GQL endpoint from another PR.

PR #2184, opened 2023-08-09
[WIP] Add query/subscription to query transaction with action type filter

XxshiftxX:
- Add ncTransactions query that can query transactions with action type filter
- Add tx subscription emitting transaction, txResults having specific action type

## Example (NOT FIXED NAMING for fields, arguments)

```graphql
subscription {
  tx {
    transaction {
		id
	}
	transactionResult {
		blockHash
		blockIndex
		txSuccess
	}
  }
}
```

```graphql
query {
	transaction {
		ncTransactions(startingBlockIndex, limit, actionType) {
			id
			action
		}
	}
}
```

### DIFF
--- a/NineChronicles.Headless/GraphTypes/TransactionHeadlessQuery.cs
+++ b/NineChronicles.Headless/GraphTypes/TransactionHeadlessQuery.cs
@@ -66,26 +66,7 @@ namespace NineChronicles.Headless.GraphTypes
                 }
             );
 
-            IEnumerable<Block> ListBlocks(BlockChain chain, long from, long limit)
-            {
-                if (chain.Count <= from)
-                {
-                    yield break;
-                }
 
-                var block = chain[from];
-                for (var i = 0; i < limit; i++)
-                {
-                    if (block is null)
-                    {
-                        yield break;
-                    }
-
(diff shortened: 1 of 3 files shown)


NineChronicles.Headless/GraphTypes/TransactionHeadlessQuery.cs
```diff
@@ -59,6 +66,65 @@ public TransactionHeadlessQuery(StandaloneContext standaloneContext)
...
+                        .Where((tx) => tx.Actions.Any(rawAction =>
+                        {
+                            if (rawAction is not Dictionary action || action["type_id"] is not Text typeId)
+                            {
+                                return false;
+                            }
+
+                            return typeId == actionType;
+                        }));
+
```
longfin, line 123: I think that we can leverage [PLINQ](https://learn.microsoft.com/en-us/dotnet/standard/parallel-programming/introduction-to-plinq) (or any other parallelized way) here because all jobs(i.e., search tx for given action type) against blocks seems to be isolated and don't affect other jobs.

moreal:
This pull request's commits will be included in `development` through `release/30` branch and #2195 PR.

Issue #2189, opened 2023-08-10
Add filter to GQL query for finding items


longfin:
## 🤔 Usage?

1. User uses `grinding` action
2. NCG Voucher subscribes `grinding` action
3. NCG Voucher checks the tx success or failure
4. NCG Voucher get equipment id from the `grinding` action
5. NCG Voucher fetches avatar inventory states for previous block and current block. (`stateQuery.avatar(addr).inventory.equipment`)
    1. Get itemId and id... 🤔  
    2. ? If it wants to fetch how many crystal minted, it should check `balance` for previous block and current block.

### IF the `TxSuccess.UpdatedStates` will be removed...

- ⚠ If there are multiple grinding actions for one signer in a block, it may be able to make mistake... 🤔

PR #2190, opened 2023-08-11
Add `itemSubType` argument to `InventoryType.equipments` field

moreal:
This pull request resolves https://github.com/planetarium/NineChronicles.Headless/issues/2189

image

### DIFF
--- a/NineChronicles.Headless/GraphTypes/States/Models/Item/InventoryType.cs
+++ b/NineChronicles.Headless/GraphTypes/States/Models/Item/InventoryType.cs
@@ -3,6 +3,7 @@ using System.Linq;
 using GraphQL;
 using GraphQL.Types;
 using Nekoyume.Model.Item;
+using NineChronicles.Headless.GraphTypes.States.Models.Item.Enum;
 
 namespace NineChronicles.Headless.GraphTypes.States.Models.Item
 {
@@ -26,15 +27,38 @@ namespace NineChronicles.Headless.GraphTypes.States.Models.Item
                     {
                         Name = "equipped",
                         Description = "filter equipped inventory item"
+                    },
+                    new QueryArgument<ItemSubTypeEnumType>
+                    {
+                        Name = "itemSubType",
+                        Description = "An item subtype for fetching only equipment where " +
+                                      "its subtype is the same. If it wasn't given, you'll " +
+                                      "get all equipment without relationship to the subtype."
                     }),
                 resolve: context =>
                 {
                     var equipments = context.Source.Equipments;
                     var equippedFilter = context.GetArgument<bool?>("equipped");
+                    var itemSubTypeFilter = context.GetArgument<ItemSubType?>("itemSubType");
                     if (equippedFilter.HasValue)
                     {
                         equipments = equipments.Where(x => x.equipped == equippedFilter.Value).ToList();
                     }
+
+                    if (itemSubTypeFilter is not null)
+                    {
+                        equipments = equipments.Where(equipment => equipment.ItemSubType == itemSubTypeFilter);
+                    }
+<<<<<<< Updated upstream
+=======
+
+                    if (itemIdsFilter is not null)
+                    {
+                        var set = itemIdsFilter.ToHashSet();
+                        equipments = equipments.Where(equipment => set.Contains(equipment.ItemId));
+                    }
+>>>>>>> Stashed changes
+
                     return equipments;
                 }
             );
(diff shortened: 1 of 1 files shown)


area363 (approved):

longfin (approved):
it looks good to me, but I think that we need additional criteria about itemId, to address #2189 completely.

moreal:
> it looks good to me, but I think that we need additional criteria about itemId, to address #2189 completely.

@longfin I couldn't imagine how they would use such APIs. I thought they can get only `Aura` subtype equipment with this API and other things can be fetched as fields (e.g., `itemId`)

Well did you imagine an API like the below one? 🤔 

```graphql
stateQuery {
	avatar(...) {
		inventory {
			equipments(itemIds: [ITEM_IDS])
		}
	}
}
```

longfin:
> > it looks good to me, but I think that we need additional criteria about itemId, to address #2189 completely.
> 
> @longfin I couldn't imagine how they would use such APIs. I thought they can get only `Aura` subtype equipment with this API and other things can be fetched as fields (e.g., `itemId`)
> 
> Well did you imagine an API like the below one? 🤔
> 
> ```graphql
> stateQuery {
> 	avatar(...) {
> 		inventory {
> 			equipments(itemIds: [ITEM_IDS])
> 		}
> 	}
> }
> ```

Right. or, do we need a new query about inventory to add filters on `equipments`?

moreal:
> Right. or, do we need a new query about inventory to add filters on `equipments`?

No. It will not need a new query. It can be just added. Then... please check the below cases.

1. There are items (A, B, C, D, E, F), and the `itemIds` argument was given with `[G]`
2. It should return nothing.

----

1. There are items (A[Armor], B[Sword], C[Ring], D[Armor], E[Ring], F[Ring]). The `itemIds` argument was given with `[A, F]`, and the `itemSubType` argument was given with `Aura`
2. It should return nothing. 

I realized why I feel a sense of incompatibility. I think that it means "I want to see the equipment" to give specific `itemIds`. Then other filters (`itemSubType`, `equipped`) seem unfamiliar.

What do you think about this? @longfin

longfin:
> I realized why I feel a sense of incompatibility. I think that it means "I want to see the equipment" to give specific `itemIds`. Then other filters (`itemSubType`, `equipped`) seem unfamiliar.

Well, as with the current implementation, I guess that and-style filter might be okay. 

i.e., it's fine.

> 1. There are items (A[Armor], B[Sword], C[Ring], D[Armor], E[Ring], F[Ring]). The itemIds argument was given with [A, F], and the itemSubType argument was given with Aura
> 2. It should return nothing.

From the context of griding detection, external service will act as like as below.

```mermaid
sequenceDiagram
    actor service
    actor headless
    
    headless -->> service: Emit griding tx event
    service -->> headless: Query item info w/ equipment ids
```

When querying item info, an additional step doesn't be needed since [`Griding` action already has ids](https://github.com/planetarium/lib9c/blob/v200050/Lib9c/Action/Grinding.cs#L29)

moreal:
@longfin I applied your idea, please review again 🙏🏻

NineChronicles.Headless/GraphTypes/States/Models/Item/InventoryType.cs
```diff
@@ -26,15 +28,42 @@ public InventoryType()
...
+                    var itemIdsFilter = context.GetArgument<Guid[]?>("itemIds");
                     if (equippedFilter.HasValue)
                     {
                         equipments = equipments.Where(x => x.equipped == equippedFilter.Value).ToList();
                     }
+
+                    if (itemSubTypeFilter.HasValue)
+                    {
+                        equipments = equipments.Where(equipment => equipment.ItemSubType == itemSubTypeFilter.Value);
+                    }
```
longfin, line 59:
```suggestion
                    if (itemSubTypeFilter is not null)
                    {
                        equipments = equipments.Where(equipment => equipment.ItemSubType == itemSubTypeFilter);
                    }
```

longfin (approved):
I left a comment related to style-consistency, but overall it seems fine.

NineChronicles.Headless/GraphTypes/States/Models/Item/InventoryType.cs
moreal: I applied it ✅ https://github.com/planetarium/NineChronicles.Headless/compare/80f4bb55d4c53832f65b5f2e47c0ec4e3396be02..98754d8e95417044ee290db1599ea8aa32d3961e

PR #2193, opened 2023-08-13
disable buffering

area363:
This PR turns off redundant buffering that leads ASPNETCORE to store excessive temporary files(
https://stackoverflow.com/questions/62910894/asp-net-core-writes-request-body-to-windows-temp-folder), a situation that has been responsible for the frequent triggering of `OOMKilled` events in the Docker container (The fix is already applied in the mainnet).


moreal (approved):

area363:
Since this is a minor fix, I will merge once the CI jobs are complete.

PR #2217, opened 2023-08-28
add account management feature

area363:
This feature manages multiple accounts that originate from the same source. The feature is turned off by default and you can turn it on by adjusting the [appsettings.json](https://github.com/planetarium/NineChronicles.Headless/compare/development...area363:account-management?expand=1#diff-e47fc585cf3ec48a24eabba8519cf4fcffbfc15f80cc907dec2379af8ebdf134R125-R130) or by adding the configuration as environment variables.

### Basic configuration
- `EnableManaging`: turn `on/off` account management mode
- `ThresholdCount`: number of accounts from a single source to start managing (eg. `50` means that if there are `50` accounts that are from a single origin, the feature will start managing the associated accounts).
- `TxIntervalMinutes`: tx interval for accounts that are under management. (e.g. `10` means that the accounts under managment are allowed 1 tx every `10` minutes).
- "ManagementTimeMinutes": the amount of time the node will not except an account's tx if it goes over the `TxIntervalMinutes` value (e.g. `10` means that if an account under managment exceeds the `TxIntervalMinutes`, the node will not stage that account's tx for `10` minutes).

### DIFF
--- a/NineChronicles.Headless/Middleware/HttpCaptureMiddleware.cs
+++ b/NineChronicles.Headless/Middleware/HttpCaptureMiddleware.cs
@@ -125,8 +125,8 @@ namespace NineChronicles.Headless.Middleware
                                         {
                                             _logger.Information($"[GRAPHQL-REQUEST-CAPTURE] Restoring Agent {agent} after {_options.Value.ManagementTimeMinutes} minutes.");
                                             RestoreMultiAccount(agent);
-                                            MultiAccountTxIntervalTracker[agent] = DateTimeOffset.Now.AddMinutes((int)-_options.Value.TxIntervalMinutes!);
-                                            _logger.Information($"[GRAPHQL-REQUEST-CAPTURE] Current time: {DateTimeOffset.Now} Added time: {DateTimeOffset.Now.AddMinutes((int)-_options.Value.TxIntervalMinutes!)}.");
+                                            MultiAccountTxIntervalTracker[agent] = DateTimeOffset.Now.AddMinutes(-_options.Value.TxIntervalMinutes);
+                                            _logger.Information($"[GRAPHQL-REQUEST-CAPTURE] Current time: {DateTimeOffset.Now} Added time: {DateTimeOffset.Now.AddMinutes(-_options.Value.TxIntervalMinutes)}.");
                                         }
                                         else
                                         {
--- a/NineChronicles.Headless/Middleware/HttpCaptureMiddleware.cs
+++ b/NineChronicles.Headless/Middleware/HttpCaptureMiddleware.cs
@@ -1,6 +1,14 @@
+using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
+using Libplanet.Common;
+using Libplanet.Crypto;
+using Libplanet.Types.Tx;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Options;
+using NineChronicles.Headless.Properties;
 using Serilog;
 using ILogger = Serilog.ILogger;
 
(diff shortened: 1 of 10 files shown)


NineChronicles.Headless/Middleware/HttpCaptureMiddleware.cs
```diff
@@ -1,20 +1,50 @@
...
+        private StandaloneContext _standaloneContext;
+        private readonly Dictionary<string, HashSet<Address>> _ipSignerList;
+        private readonly IOptions<MultiAccountManagerProperties> _options;
 
-        public HttpCaptureMiddleware(RequestDelegate next)
+        public HttpCaptureMiddleware(
+            RequestDelegate next,
+            StandaloneContext standaloneContext,
+            Dictionary<string, HashSet<Address>> ipSignerList,
+            IOptions<MultiAccountManagerProperties> options)
```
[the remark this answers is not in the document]
area363: Sure, I'll separate this middleware along with the one mentioned in https://github.com/planetarium/NineChronicles.Headless/pull/2217#discussion_r1308014604.

area363:
@moreal I've separated the middlewares and rebased this PR to `release/40`. PTAL! 🙏

NineChronicles.Headless/Middleware/HttpMultiAccountManagementMiddleware.cs
```diff
@@ -0,0 +1,185 @@
...
+                            UpdateIpSignerList(remoteIp!.ToString(), agent);
+                        }
+                        catch (Exception ex)
+                        {
+                            _logger.Error(
+                                "[GRAPHQL-MULTI-ACCOUNT-MANAGER] Error message: {message} Stacktrace: {stackTrace}",
+                                ex.Message,
+                                ex.StackTrace);
+                        }
+                    }
```
moreal, line 75: Not critical to merge but it seems better to parse with regex or real GraphQL parser 🤔

NineChronicles.Headless/Middleware/HttpMultiAccountManagementMiddleware.cs
```diff
@@ -0,0 +1,185 @@
...
+                                        {
+                                            _logger.Information($"[GRAPHQL-MULTI-ACCOUNT-MANAGER] Restoring Agent {agent} after {_options.Value.ManagementTimeMinutes} minutes.");
+                                            RestoreMultiAccount(agent);
+                                            MultiAccountTxIntervalTracker[agent] = DateTimeOffset.Now.AddMinutes(-_options.Value.TxIntervalMinutes);
+                                            _logger.Information($"[GRAPHQL-MULTI-ACCOUNT-MANAGER] Current time: {DateTimeOffset.Now} Added time: {DateTimeOffset.Now.AddMinutes(-_options.Value.TxIntervalMinutes)}.");
+                                        }
+                                        else
+                                        {
+                                            _logger.Information($"[GRAPHQL-MULTI-ACCOUNT-MANAGER] Agent {agent} is in managed status for the next {_options.Value.ManagementTimeMinutes - (DateTimeOffset.Now - MultiAccountManagementList[agent]).Minutes} minutes.");
```
moreal, line 131: It seems better to extract `(DateTimeOffset.Now - MultiAccountManagementList[agent]).Minutes` as a new variable.

moreal (approved):

NineChronicles.Headless/Middleware/HttpMultiAccountManagementMiddleware.cs
area363: That sounds better. Let me look into this 👍

area363: @moreal I've looked into this and I was able to parse the request by doing something like this:

```
public class GraphQlRequest
{
    public string? Query { get; set; }
    public Dictionary<string, object>? Variables { get; set; }
}

GraphQlRequest graphQLRequest = JsonConvert.DeserializeObject<GraphQlRequest>(body);
```

However, the problem is, `JsonConvert.DeserializeObject()` takes alittle long to deserialize and I'm worried that it'll affect the overall response performance 😢.

I'll merge this for now and create an new issue to address a more efficient way to parse the data. 🙏

moreal: Okay! ✅

moreal (approved):

PR #2238, opened 2023-09-06
Introduce auraSummon ActionQuery

U-lis:
SSIA


NineChronicles.Headless/GraphTypes/ActionQueryFields/Summon.cs
moreal, line 1: Will there be other `Summon` series actions? If not, I think the filename should be `AuraSummon`

U-lis: Yes, another `*Summon` action series are coming sooner or later.
I'm planning to put all summon series to single `Summon.cs` file.

moreal (approved):

PR #2242, opened 2023-09-12
fix duplicate publisher in RPC mode

area363:
This PR addresses the issue where `ActionEvaluationPublisher` was instantiated twice in RPC mode. Now, the publisher is initialized in `Program.cs` and subsequently passed to both `UseNineChroniclesNode` and `UseNineChroniclesNodeRPC` methods found in `HostBuilderExtensions.cs`.


area363:
@riemannulus Could you take a look at this? I've tested it on the mainnet and it works as expected. 🙏

riemannulus (approved):

PR #2246, opened 2023-09-14
✨ implement: `IBlockChainService.GetBlockHash`

riemannulus:
Implement `IBlockChainService.GetBlockHash` to get `BlockHash` from `BlockIndex` with gRPC.

### DIFF
--- a/NineChronicles.Headless/BlockChainService.cs
+++ b/NineChronicles.Headless/BlockChainService.cs
@@ -180,6 +180,18 @@ namespace NineChronicles.Headless
             return new UnaryResult<byte[]>(headerBytes);
         }
 
+        public UnaryResult<byte[]> GetBlockHash(long blockIndex)
+        {
+            try
+            {
+                return new UnaryResult<byte[]>(_codec.Encode(_blockChain[blockIndex].Hash.Bencoded));
+            }
+            catch (ArgumentOutOfRangeException)
+            {
+                return new UnaryResult<byte[]>(_codec.Encode(new Null()));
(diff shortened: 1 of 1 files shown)


NineChronicles.Headless/BlockChainService.cs
```diff
@@ -180,6 +180,18 @@ public UnaryResult<byte[]> GetTip()
...
 
+        public UnaryResult<byte[]> GetBlockHash(long blockIndex)
+        {
+            try
+            {
+                return new UnaryResult<byte[]>(_codec.Encode(_blockChain[blockIndex].Hash.Bencoded));
+            }
+            catch (ArgumentOutOfRangeException)
+            {
+                return new UnaryResult<byte[]>(_codec.Encode(new Null()));
```
moreal, line 191: How about using static `Null` instance?

https://github.com/planetarium/bencodex.net/blob/c707249e235181f52c2aef5d4e010390ce5ac894/Bencodex/Types/Null.cs#L13-L20

```suggestion
                return new UnaryResult<byte[]>(_codec.Encode(Null.Value));
```

moreal (approved):

NineChronicles.Headless/BlockChainService.cs
riemannulus: https://github.com/planetarium/NineChronicles.Headless/pull/2246/commits/2359545ad28ad062d87092cb14a33a42bc5c468d amended!

moreal (approved):

PR #2255, opened 2023-09-21
🔧  Remove `IRenderer`s from `BlockPolicySource`

riemannulus:
SSIA


moreal (approved):
I should check gRPC features work, but it looks good at code level 👍🏻 😎

PR #2260, opened 2023-10-04
🚀  bump: `NineChronicles.RPC.Shared`

riemannulus:
SSIA

### DIFF
--- a/NineChronicles.Headless/BlockChainService.cs
+++ b/NineChronicles.Headless/BlockChainService.cs
@@ -4,10 +4,13 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
+using System.Security.Cryptography;
 using System.Threading.Tasks;
 using Bencodex;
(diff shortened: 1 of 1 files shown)


NineChronicles.Headless/BlockChainService.cs
```diff
@@ -140,6 +152,26 @@ public UnaryResult<byte[]> GetState(byte[] addressBytes, byte[] blockHashBytes)
             return result.ToDictionary(kv => kv.Key, kv => kv.Value);
         }
 
+        public async UnaryResult<Dictionary<byte[], byte[]>> GetAvatarStatesBySrh(
```
moreal, line 155: I think this method is conflicted with `GetStateBulkBySrh` method.

moreal: Ah.. It seems like a version with stateroothash of `GetAvatarStates` method that already exists.

moreal (approved):

moreal:
It needs to rebase.

PR #2271, opened 2023-10-17
Implement swagger page

Atralupus:
image
SSIA

### DIFF
--- a/NineChronicles.Headless.AccessControlCenter/AccessControlService/MutableSqliteAccessControlService.cs
+++ b/NineChronicles.Headless.AccessControlCenter/AccessControlService/MutableSqliteAccessControlService.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using Microsoft.Data.Sqlite;
 using Libplanet.Crypto;
@@ -39,6 +40,11 @@ namespace NineChronicles.Headless.AccessControlCenter.AccessControlService
 
         public List<Address> ListBlockedAddresses(int offset, int limit)
         {
+            if (limit > 30)
+            {
+                throw new ArgumentException("Limit cannot exceed 30.", nameof(limit));
+            }
+
             var blockedAddresses = new List<Address>();
 
             using var connection = new SqliteConnection(_connectionString);
--- a/NineChronicles.Headless.AccessControlCenter/AccessControlService/MutableSqliteAccessControlService.cs
+++ b/NineChronicles.Headless.AccessControlCenter/AccessControlService/MutableSqliteAccessControlService.cs
@@ -40,11 +40,6 @@ namespace NineChronicles.Headless.AccessControlCenter.AccessControlService
 
         public List<Address> ListBlockedAddresses(int offset, int limit)
         {
-            if (limit > 30)
-            {
-                throw new ArgumentException("Limit cannot exceed 30.", nameof(limit));
-            }
-
             var blockedAddresses = new List<Address>();
 
             using var connection = new SqliteConnection(_connectionString);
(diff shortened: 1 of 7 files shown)


NineChronicles.Headless.AccessControlCenter/AccessControlService/MutableSqliteAccessControlService.cs
```diff
@@ -39,6 +40,11 @@ public void AllowAccess(Address address)
 
         public List<Address> ListBlockedAddresses(int offset, int limit)
         {
+            if (limit > 30)
```
area363, line 43: Is there a particular reason why the limit is set to 30?

NineChronicles.Headless.AccessControlCenter/Controllers/AccessControlServiceController.cs
```diff
@@ -39,10 +40,17 @@ public ActionResult AllowAccess(string address)
...
+            try
+            {
+                return _accessControlService
+                    .ListBlockedAddresses(offset, limit)
+                    .Select(a => a.ToString())
+                    .ToList();
+            }
+            catch (ArgumentException ex)
+            {
+                return BadRequest(ex.Message);
```
moreal, line 52: It seems to be `InternalServerError` instead of `BadRequest` or I think the `AccessConrolServiceController.ListBlockedAddresses(...)` should check the `limit` variable manually.

NineChronicles.Headless.AccessControlCenter/AccessControlService/MutableSqliteAccessControlService.cs
Atralupus: There's no reason for sqlite to have such a low limit...
Do you think a maximum of 100 is acceptable?

NineChronicles.Headless.AccessControlCenter/Controllers/AccessControlServiceController.cs
Atralupus: Both ACS service have different  limit values.
Is this a proper way to implement limit validation?
```csharp
AccessControlServiceController.cs > ListBlockedAddresses
...
            var maxLimit = 10;
            if (_accessControlService.GetType() == typeof(MutableRedisAccessControlService))
            {
                maxLimit = 10;
            }
            else if (_accessControlService.GetType() == typeof(MutableSqliteAccessControlService))
            {
                maxLimit = 100;
            }

            if (limit > maxLimit)
            {
                throw new ArgumentException($"The limit cannot exceed {maxLimit}.", nameof(limit));
            }
```

moreal: The reason I thought it was necessary to validate directly at the controller level is because `ArgumentException` is such a common error. It can also happen in the constructor of `Address` if Redis has an unintended invalid key in it, which is usually not the case. One way to do this would be to **(1) create a separate, dedicated exception that inherits from `ArgumentException`.** Or, **(2) provide a value for Limit as a read-only property in the `IMutableAccessControlService` interface, and use it.** Or, **(3) go with something like the code you posted, and use the `is` keyword to check for type.**

Atralupus: How about this?

```
        [HttpGet("entries")]
        public ActionResult<List<string>> ListBlockedAddresses(int offset, int limit)
        {
            var maxLimit = 10;
            if (_accessControlService is MutableRedisAccessControlService)
            {
                maxLimit = 10;
            }
            else if (_accessControlService is MutableSqliteAccessControlService)
            {
                maxLimit = 100;
            }

            if (limit > maxLimit)
            {
                return BadRequest($"The limit cannot exceed {maxLimit}.");
            }

            return _accessControlService
                .ListBlockedAddresses(offset, limit)
                .Select(a => a.ToString())
                .ToList();
        }
```
implement option 3

moreal: LGTM 👍🏻

moreal (approved):
I'll fix little details in other pr

PR #2276, opened 2023-10-26
Add whitelist option to ACC service

area363:
Previously, the ACC checked if an address simply exists in the db to determine whether it should be denied.

The new update refers to the value of the address key to determine its 'access level":

- A value of `0` means that the address should be denied.
- A value of `1` means that the address is whitelisted and all txs should be staged without limit.

In the future, the numeric value of the key will be used to allow more txs to be mined in a block (This change needs further investigation into dynamically adjusting the `BlockPolicy`).

p.s. This PR requires a new `lib9c` bump after https://github.com/planetarium/lib9c/pull/2191 is merged.


area363:
Replaced by https://github.com/planetarium/NineChronicles.Headless/pull/2278.

PR #2278, opened 2023-10-31
Add `AddTxQuota` and `RemoveTxQuota` to ACS

area363:
### Changes
- Add `AddTxQuota` and `RemoveTxQuota` to ACS to set or delete individual signer's `txQuotaPerSigner`.

Replaces https://github.com/planetarium/NineChronicles.Headless/pull/2276.

### DIFF
--- a/NineChronicles.Headless.AccessControlCenter/AccessControlService/MutableRedisAccessControlService.cs
+++ b/NineChronicles.Headless.AccessControlCenter/AccessControlService/MutableRedisAccessControlService.cs
@@ -1,4 +1,3 @@
-using System;
 using System.Collections.Generic;
 using System.Linq;
 using Libplanet.Crypto;
@@ -16,20 +15,6 @@ namespace NineChronicles.Headless.AccessControlCenter.AccessControlService
         {
         }
 
-        public void DenyAccess(Address address)
-        {
-            _db.StringSet(address.ToString(), "0");
-        }
-
-        public void AllowAccess(Address address)
-        {
-            var value = _db.StringGet(address.ToString());
-            if (value == "0")
-            {
-                _db.KeyDelete(address.ToString());
-            }
-        }
-
         public void AddTxQuota(Address address, int quota)
         {
             _db.StringSet(address.ToString(), quota.ToString());
@@ -40,7 +25,7 @@ namespace NineChronicles.Headless.AccessControlCenter.AccessControlService
             _db.KeyDelete(address.ToString());
         }
 
-        public List<Address> ListBlockedAddresses(int offset, int limit)
+        public List<Address> ListTxQuotaAddresses(int offset, int limit)
         {
             var server = _db.Multiplexer.GetServer(_db.Multiplexer.GetEndPoints().First());
 
@@ -48,11 +33,10 @@ namespace NineChronicles.Headless.AccessControlCenter.AccessControlService
                 server.Execute("SCAN", offset.ToString(), "COUNT", limit.ToString());
             if (result != null)
             {
-                long newCursor = long.Parse((string)result[0]!);
                 RedisKey[] keys = (RedisKey[])result[1]!;
-
                 return keys.Select(k => new Address(k.ToString())).ToList();
             }
+
             return new List<Address>();
         }
     }
--- a/NineChronicles.Headless.AccessControlCenter/AccessControlService/MutableRedisAccessControlService.cs
+++ b/NineChronicles.Headless.AccessControlCenter/AccessControlService/MutableRedisAccessControlService.cs
@@ -18,10 +18,24 @@ namespace NineChronicles.Headless.AccessControlCenter.AccessControlService
 
         public void DenyAccess(Address address)
         {
-            _db.StringSet(address.ToString(), "denied");
+            _db.StringSet(address.ToString(), "0");
         }
 
         public void AllowAccess(Address address)
+        {
+            var value = _db.StringGet(address.ToString());
+            if (value == "0")
+            {
+                _db.KeyDelete(address.ToString());
+            }
+        }
+
+        public void AddTxQuota(Address address, int quota)
+        {
+            _db.StringSet(address.ToString(), quota.ToString());
+        }
+
+        public void RemoveTxQuota(Address address)
         {
             _db.KeyDelete(address.ToString());
         }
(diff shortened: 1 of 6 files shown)


NineChronicles.Headless.AccessControlCenter/AccessControlService/MutableRedisAccessControlService.cs
```diff
@@ -18,10 +18,24 @@ public MutableRedisAccessControlService(string storageUri)
 
         public void DenyAccess(Address address)
         {
-            _db.StringSet(address.ToString(), "denied");
```
moreal, line 21: It is not a service running somewhere, right? If it is running, it should care for backward-compatibility

NineChronicles.Headless.AccessControlCenter/AccessControlService/MutableRedisAccessControlService.cs
```diff
@@ -18,10 +18,24 @@ public MutableRedisAccessControlService(string storageUri)
...
+            _db.StringSet(address.ToString(), "0");
         }
 
         public void AllowAccess(Address address)
+        {
+            var value = _db.StringGet(address.ToString());
+            if (value == "0")
+            {
+                _db.KeyDelete(address.ToString());
+            }
```
moreal, line 30: It seems to use the same key for several purposes. I think it is an easy way to cause bugs. How about using different keys with prefixes? If it removes tx quota, it will be `access-denined` state.

NineChronicles.Headless.AccessControlCenter/AccessControlService/MutableSqliteAccessControlService.cs
```diff
@@ -9,9 +9,13 @@ namespace NineChronicles.Headless.AccessControlCenter.AccessControlService
     public class MutableSqliteAccessControlService : SQLiteAccessControlService, IMutableAccessControlService
     {
         private const string DenyAccessSql =
-            "INSERT OR IGNORE INTO blocklist (address) VALUES (@Address)";
+            "INSERT OR IGNORE INTO blocklist (address, quota) VALUES (@Address, 0)";
         private const string AllowAccessSql = "DELETE FROM blocklist WHERE address=@Address";
 
+        private const string AddTxQuotaSql =
+            "INSERT OR IGNORE INTO blocklist (address, quota) VALUES (@Address, @Quota)";
+        private const string RemoveTxQuotaSql = "DELETE FROM blocklist WHERE address=@Address";
```
moreal, line 17: How about using a separated table? Please see the mention I left above.

NineChronicles.Headless.AccessControlCenter/Controllers/AccessControlServiceController.cs
```diff
@@ -36,6 +37,21 @@ public ActionResult AllowAccess(string address)
             return Ok();
         }
 
+        [HttpPost("entries/add-tx-quota/{address}/{quota}")]
```
moreal, line 40:
```suggestion
        [HttpPost("entries/add-tx-quota/{address}/{quota:int}")]
```

Reference: https://learn.microsoft.com/en-us/aspnet/core/mvc/controllers/routing?view=aspnetcore-7.0#attribute-routing-with-http-verb-attributes

NineChronicles.Headless.AccessControlCenter/AccessControlService/MutableRedisAccessControlService.cs
area363: @moreal Thanks for the review!

I've just had a discussion with @Atralupus and we've decided to remove `AllowAccess` and `DenyAccess` completely because the function of the `ACS` has been changed from "controlling access" to "managing tx quotas". I'll make these new updates along with your other suggestion and request one more review. 🙏

area363:
@moreal As mentioned in https://github.com/planetarium/NineChronicles.Headless/pull/2278#discussion_r1378417189, I've made further changes to the PR. 

### New Changes

- `DenyAccess` and `AllowAccess` have been completely removed via @Atralupus 's suggestion and only `AddTxQuota` and `RemoveTxQuota` are used.
- `IsAccessDenied` has been replaced with `IsListed`, to check if an address is listed in the DB with a tx quota value.

PTAL! 🙏

NineChronicles.Headless.AccessControlCenter/AccessControlService/MutableRedisAccessControlService.cs
```diff
@@ -16,29 +15,28 @@ public MutableRedisAccessControlService(string storageUri)
         {
         }
 
-        public void DenyAccess(Address address)
+        public void AddTxQuota(Address address, int quota)
```
Atralupus, line 18: I think we need to set a quota limit, perhaps... at 10?

area363: @Atralupus Should I add the same limitation(https://github.com/planetarium/NineChronicles.Headless/pull/2278/commits/7d3a51adca1d9bf5383d26f6a534c9852a2948d0) here?

Atralupus: yes, right

NineChronicles.Headless.AccessControlCenter/Controllers/AccessControlServiceController.cs
```diff
@@ -17,22 +18,28 @@ public AccessControlServiceController(IMutableAccessControlService accessControl
...
-        [HttpPost("entries/{address}/deny")]
-        public ActionResult DenyAccess(string address)
+        [HttpPost("entries/add-tx-quota/{address}/{quota:int}")]
+        public ActionResult AddTxQuota(string address, int quota)
         {
-            _accessControlService.DenyAccess(new Address(address));
+            var maxLimit = 10;
+            if (quota > maxLimit)
+            {
+                return BadRequest($"The limit cannot exceed {maxLimit}.");
```
moreal, line 32: It isn't critical but I think it is good to unify terms. How about using `quota` instead of `limit`?

moreal (approved):

Atralupus (approved):

NineChronicles.Headless.AccessControlCenter/Controllers/AccessControlServiceController.cs
area363: @moreal You mean `maxLimit` to `maxQuota` right?

moreal: Ah, it is a good point too. I mean the message in the `BadRequest` function's argument because it is shown to users.
```suggestion
                return BadRequest($"The quota cannot exceed {maxLimit}.");
```

area363: Ah i see it now. I'll unify the terms, thank you!

moreal (approved):
👍🏻

PR #2281, opened 2023-11-01
Tweak genesis command 

longfin:
This PR aims to tweak genesis command like as below:

- Drops activation related features.
- Adds `initialMeadConfigs` and `initialPledgeConfigs`.
- Adds `currency.allowMint` for minter-less major currency.


moreal (approved):
If you don't care about the breaking change, it seems okay.

PR #2282, opened 2023-11-02
Introduce ClaimItems GQL endpoint

U-lis:
SSIA

### DIFF
--- a/NineChronicles.Headless.Tests/GraphTypes/ActionQueryTest.cs
+++ b/NineChronicles.Headless.Tests/GraphTypes/ActionQueryTest.cs
@@ -24,7 +24,6 @@ using Nekoyume.Model.State;
 using Nekoyume.TableData;
 using NineChronicles.Headless.GraphTypes;
 using Xunit;
-using Xunit.Abstractions;
 using static NineChronicles.Headless.Tests.GraphQLTestUtils;
 
 namespace NineChronicles.Headless.Tests.GraphTypes
@@ -1346,6 +1345,8 @@ actionPoint: {actionPoint},
         [Theory]
         [InlineData(1)]
         [InlineData(2)]
+        [InlineData(10)]
+        [InlineData(100)]
         public async Task ClaimItems(int claimDataCount)
         {
             var random = new Random();
@@ -1366,7 +1367,7 @@ actionPoint: {actionPoint},
                         Currency.Uncapped(
                             ticker: ticker,
                             decimalPlaces: 0,
-                            minters: new AddressSet(new List<Address>())
+                            minters: AddressSet.Empty
                         ),
                         random.Next(1, 100), 0
                     )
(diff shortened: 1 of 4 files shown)


boscohyun (approved):

NineChronicles.Headless/GraphTypes/Input/ClaimDataInputType.cs
```diff
@@ -0,0 +1,25 @@
+using System.Collections.Generic;
+using System.Linq;
+using GraphQL.Types;
+using Libplanet.Crypto;
+using Libplanet.Explorer.GraphTypes;
+using Libplanet.Types.Assets;
+
+namespace NineChronicles.Headless.GraphTypes.Input;
+
+public class ClaimDataInputType : InputObjectGraphType<(Address avatarAddress, FungibleAssetValue fav)>
```
moreal, line 10: Maybe it should type `IReadOnlyList<FungibleAssetValue> favs` or implement `InputObjectGraphType` (non-generic) instead. I'm not sure why the second element type is just `FungibleAssetValue` 👀

NineChronicles.Headless.Tests/GraphTypes/ActionQueryTest.cs
```diff
@@ -1340,5 +1342,81 @@ public async Task AuraSummon()
...
+                queryBuilder.Append($@"{{
+                    avatarAddress: ""{avatarAddress}"",
+                    fungibleAssetValues:[
+                ");
+                var favList = tickerCandidates.Select(
+                    ticker => new FungibleAssetValue(
+                        Currency.Uncapped(
+                            ticker: ticker,
+                            decimalPlaces: 0,
+                            minters: new AddressSet(new List<Address>())
```
moreal, line 1369:
```suggestion
                            minters: AddressSet.Empty
```

NineChronicles.Headless.Tests/GraphTypes/ActionQueryTest.cs
```diff
@@ -1340,5 +1342,81 @@ public async Task AuraSummon()
...
+
+            for (var i = 0; i < claimDataList.Count; i++)
+            {
+                var (expectedAddr, expectedFavList) = claimDataList[i];
+                var (actualAddr, actualFavList) = action.ClaimData[i];
+                Assert.Equal(expectedAddr, actualAddr);
+                Assert.Equal(expectedFavList.Count, actualFavList.Count);
+                for (var j = 0; j < expectedFavList.Count; j++)
+                {
+                    // Assert.Equal(expectedFavList[i], actualFavList[i]);
```
moreal, line 1415: Why is this line disabled? (by comment)

U-lis: That is not working because Currency hash of two FAVs are different.
I found the reason was different minters, I provided `[]` but currency from action has `null`
I tried to give null as a minters parameter but GQL does not allow this. So I gave up to fix this :fearful:

moreal: Yes, it is GraphType's bug... 🐛  I think I can fix it easily but I'm scared if it makes a huge breaking-change. At this time, I understood why the line is disabled.

U-lis: I'll add additional comment on top of the this line why this is disabled and how to fix this :+1:

moreal (approved):

PR #2304, opened 2023-11-15
update acs log level to verbose

area363:
SSIA


area363:
Merging because it's a minor log level change.

PR #2297, opened 2023-11-09
set ACS redis timeout to 500ms

area363:
Fix the following issue:

```
[01:59:26 ERR] StackExchange.Redis.RedisTimeoutException: Timeout performing GET (5000ms), next: GET 0xE86729DB3468a5676D7BAbC67fA444fe1F9E0175, inst: 2, qu: 0, qs: 10, aw: False, bw: SpinningDown, rs: ReadAsync, ws: Idle, in: 50, in-pipe: 0, out-pipe: 0, last-in: 0, cur-in: 0, sync-ops: 31303, async-ops: 0, serverEndpoint: acc-redis-service.9c-network.svc.cluster.local:6379, conn-sec: 1093.82, mc: 1/1/0, mgr: 10 of 10 available, clientName: remote-headless-2-0(SE.Redis-v2.6.96.30123), IOCP: (Busy=0,Free=1000,Min=8,Max=1000), WORKER: (Busy=66,Free=32701,Min=8,Max=32767), POOL: (Threads=83,QueuedItems=546,CompletedItems=1083658), v: 2.6.96.30123 (Please take a look at this article for some common client-side issues that can cause timeouts: https://stackexchange.github.io/StackExchange.Redis/Timeouts)
   at StackExchange.Redis.ConnectionMultiplexer.ExecuteSyncImpl[T](Message message, ResultProcessor`1 processor, ServerEndPoint server, T defaultValue) in /_/src/StackExchange.Redis/ConnectionMultiplexer.cs:line 1994
   at StackExchange.Redis.RedisBase.ExecuteSync[T](Message message, ResultProcessor`1 processor, ServerEndPoint server, T defaultValue) in /_/src/StackExchange.Redis/RedisBase.cs:line 62
   at StackExchange.Redis.RedisDatabase.StringGet(RedisKey key, CommandFlags flags) in /_/src/StackExchange.Redis/RedisDatabase.cs:line 2995
   at NineChronicles.Headless.Services.RedisAccessControlService.GetTxQuota(Address address) in /app/NineChronicles.Headless/Services/RedisAccessControlService.cs:line 26
   at Nekoyume.Blockchain.NCStagePolicy.Iterate(BlockChain blockChain, Boolean filtered) in /app/Lib9c/Lib9c.Policy/NCStagePolicy.cs:line 51
   at Libplanet.Blockchain.BlockChain.GetStagedTransactionIds() in /app/Lib9c/.Libplanet/Libplanet/Blockchain/BlockChain.cs:line 713
[...]

PR #2306, opened 2023-11-17
set timeout to 0.5s

area363:
Replaces https://github.com/planetarium/NineChronicles.Headless/pull/2297.


moreal (approved):
In the future, it should be configurable.

PR #2318, opened 2023-11-21
detailed log arena list worker

ipdae:
more detailed worker log for arena list caching

### DIFF
--- a/NineChronicles.Headless/ArenaParticipantsWorker.cs
+++ b/NineChronicles.Headless/ArenaParticipantsWorker.cs
@@ -52,16 +52,18 @@ public class ArenaParticipantsWorker : BackgroundService
         catch (OperationCanceledException)
         {
             //pass
+            _logger.Information("[ArenaParticipantsWorker]Cancel ArenaParticipantsWorker");
         }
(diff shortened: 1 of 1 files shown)


moreal (approved):
`ForContext` is probably the way to go when filtering in Loki. I don't know exactly, so it would be nice if Kidon could talk about it.

https://github.com/planetarium/NineChronicles.Headless/blob/e89c6a9e296a65f893c3c049654a155220de97af/NineChronicles.Headless/Middleware/CustomRateLimitMiddleware.cs#L27

area363:
> `ForContext` is probably the way to go when filtering in Loki. I don't know exactly, so it would be nice if @area363 could talk about it.
> 
> https://github.com/planetarium/NineChronicles.Headless/blob/e89c6a9e296a65f893c3c049654a155220de97af/NineChronicles.Headless/Middleware/CustomRateLimitMiddleware.cs#L27

@moreal Yes `ForContext` is better for Loki parsing. But it's not critical so I think we can update this part in the future if necessary. cc @ipdae

area363 (approved):

PR #2326, opened 2023-11-27
Introduce TransactionResults query

ipdae:
introduce bulk api for transaction result check from #2323

### DIFF
--- a/NineChronicles.Headless/GraphTypes/TransactionHeadlessQuery.cs
+++ b/NineChronicles.Headless/GraphTypes/TransactionHeadlessQuery.cs
@@ -192,42 +192,25 @@ namespace NineChronicles.Headless.GraphTypes
                 ),
                 resolve: context =>
                 {
-                    if (!(standaloneContext.BlockChain is BlockChain blockChain))
-                    {
-                        throw new ExecutionError(
-                            $"{nameof(StandaloneContext)}.{nameof(StandaloneContext.BlockChain)} was not set yet!");
-                    }
+                    var txId = context.GetArgument<TxId>("txId");
+                    return TxResult(standaloneContext, txId);
+                });
 
-                    if (!(standaloneContext.Store is IStore store))
+            Field<NonNullGraphType<ListGraphType<TxResultType>>>(
+                name: "transactionResults",
+                arguments: new QueryArguments(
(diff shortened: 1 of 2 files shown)


NineChronicles.Headless/GraphTypes/TransactionHeadlessQuery.cs
```diff
@@ -312,6 +295,45 @@ public TransactionHeadlessQuery(StandaloneContext standaloneContext)
...
+                    txExecutedBlock.Index,
+                    txExecutedBlock.Hash.ToString(),
+                    execution.InputState,
+                    execution.OutputState,
+                    execution.ExceptionNames);
+            }
+            catch (Exception)
+            {
+                return new TxResult(TxStatus.INVALID, null, null, null, null, null);
+            }
```
moreal, line 334: When the exception will be raised? I think the exception type should be specified more (not just `Exception`). If not, it will make us hard to debug why the request failed.

NineChronicles.Headless/GraphTypes/TransactionHeadlessQuery.cs
```diff
@@ -192,42 +192,25 @@ public TransactionHeadlessQuery(StandaloneContext standaloneContext)
...
-                            txExecutedBlock.Hash.ToString(),
-                            execution.InputState,
-                            execution.OutputState,
-                            execution.ExceptionNames);
-                    }
-                    catch (Exception)
+            Field<NonNullGraphType<ListGraphType<TxResultType>>>(
+                name: "transactionResults",
+                arguments: new QueryArguments(
+                    new QueryArgument<NonNullGraphType<ListGraphType<TxIdType>>>
```
moreal, line 202:
```suggestion
                    new QueryArgument<NonNullGraphType<ListGraphType<NonNullGraphType<TxIdType>>>>
```

NineChronicles.Headless/GraphTypes/TransactionHeadlessQuery.cs
moreal: I misread the change, this part was already working like this. It would be nice to fix it separately in the next PR.

moreal (approved):

PR #2335, opened 2023-12-04
fix table caching

ipdae:
- Introduce MemoryCacheExtensions for caching sheet
- Add cache expiration time
- Introduce cached sheet query

### DIFF
--- a/NineChronicles.Headless/GraphTypes/StateQuery.cs
+++ b/NineChronicles.Headless/GraphTypes/StateQuery.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Bencodex;
 using Bencodex.Types;
 using GraphQL;
 using GraphQL.Types;
@@ -11,7 +12,6 @@ using Libplanet.Explorer.GraphTypes;
 using Nekoyume;
 using Nekoyume.Action;
 using Nekoyume.Arena;
-using Nekoyume.Battle;
 using Nekoyume.Extensions;
 using Nekoyume.Model.Arena;
 using Nekoyume.Model.EnumType;
@@ -32,6 +32,8 @@ namespace NineChronicles.Headless.GraphTypes
 {
     public partial class StateQuery : ObjectGraphType<StateContext>
     {
+        private readonly Codec _codec = new Codec();
+
         public StateQuery()
         {
             Name = "StateQuery";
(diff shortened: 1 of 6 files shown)


NineChronicles.Headless/GraphTypes/StateQuery.cs
```diff
@@ -699,6 +701,22 @@ public StateQuery()
...
+                        Name = "tableName"
+                    }
+                ),
+                resolve: context =>
+                {
+                    var tableName = context.GetArgument<string>("tableName");
+                    var cacheKey = Addresses.GetSheetAddress(tableName).ToString();
+                    return context.Source.StateMemoryCache.SheetCache.GetSheet(cacheKey);
+                }
+            );
```
moreal, line 719: Where is this API used? For debugging?

If then, though it may not be a recommended way, how about making it as a `@deprecated` query to prevent use by others?

NineChronicles.Headless/MemoryCacheExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        var compressed = MessagePackSerializer.Serialize(Codec.Encode(value), Lz4Options);
+        cache.Set(cacheKey, compressed, ex);
+        return compressed;
+    }
+
+    public static bool TryGetSheet<T>(this MemoryCache cache, string cacheKey, out T cached)
+    {
+        return cache.TryGetValue(cacheKey, out cached);
+    }
```
moreal, line 24: (This is a review you can skip if you’re in a hurry) Is there a reason to make an alias of `cache.TryGetValue`?

ipdae: After modifying the code, it no longer need it. I will delete this.

NineChronicles.Headless/GraphTypes/StateQuery.cs
ipdae: It was created to be used to compare chain and cached sheet data when operating a node.

NineChronicles.Headless/GraphTypes/StateQuery.cs
```diff
@@ -699,6 +701,22 @@ public StateQuery()
                     return result;
                 }
             );
+
+            Field<StringGraphType>(
+                name: "cachedSheet",
+                arguments: new QueryArguments(
+                    new QueryArgument<StringGraphType>
```
moreal, line 708: (It is not critical because we expect this API will be used only for internal operation)

```suggestion
                    new QueryArgument<NonNullGraphType<StringGraphType>>
```

PR #2357, opened 2023-12-13
Introduce TransferAsset JSON schema

moreal:


moreal:
This pull request will not be managed anymore.

PR #2444, opened 2024-04-17
Apply latest lib9c for rune level bonus

U-lis:
Update submodules to apply rune level bonus.
This will propagate to DataProvider.


s2quake:
This Pull Request includes fixes for errors caused by changes to nullable types in `libplanet`. While throwing exceptions is the preferred method for handling such issues, some code has used the '!' operator instead of throwing exceptions. This approach can lead to reduced readability and make it difficult for developers to identify potential issues. Therefore, for clarity and improved reliability, it is necessary to fix the code to use exception throwing instead.

s2quake (approved):

PR #2456, opened 2024-04-22
add PLUGIN_PATH env to avoid downloading the same plugin file each restart

area363:
* Adding "PLUGIN_PATH=/data" env will store plugin files in pvc and avoid re-download each restart.
* Replace GetHashCode with CreateSha256Hash to get consistent path value (GetHashCode generates random value each time).

### DIFF
--- a/Libplanet.Headless/Hosting/LibplanetNodeService.cs
+++ b/Libplanet.Headless/Hosting/LibplanetNodeService.cs
@@ -642,13 +642,13 @@ namespace Libplanet.Headless.Hosting
             if (Uri.IsWellFormedUriString(path, UriKind.Absolute))
             {
                 // Run the download on a background thread
-                return Task.Run(() => DownloadPlugin(path, SwarmCancellationToken)).GetAwaiter().GetResult();
+                return Task.Run(() => DownloadPlugin(path)).GetAwaiter().GetResult();
             }
 
             return path;
         }
 
-        private async Task<string> DownloadPlugin(string url, CancellationToken cancellationToken)
+        private async Task<string> DownloadPlugin(string url)
         {
             var basePath = Environment.GetEnvironmentVariable("PLUGIN_PATH") ?? Environment.CurrentDirectory;
             var path = Path.Combine(basePath, "plugins");
@@ -658,14 +658,16 @@ namespace Libplanet.Headless.Hosting
             using var httpClient = new HttpClient();
             var downloadPath = Path.Join(path, hashed + ".zip");
             var extractPath = Path.Join(path, hashed);
+            var checksumPath = Path.Join(extractPath, "checksum.txt");
+
             logger.Debug("Download Path: {downloadPath}", downloadPath);
             logger.Debug("Extract Path: {extractPath}", extractPath);
 
             if (!File.Exists(downloadPath))
             {
                 logger.Debug("Downloading...");
-                var content = await httpClient.GetByteArrayAsync(url, cancellationToken);
-                await File.WriteAllBytesAsync(downloadPath, content, cancellationToken);
+                var content = await httpClient.GetByteArrayAsync(url);
+                await File.WriteAllBytesAsync(downloadPath, content);
                 logger.Debug("Finished downloading.");
             }
             else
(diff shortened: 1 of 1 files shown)


Libplanet.Headless/Hosting/LibplanetNodeService.cs
```diff
@@ -635,29 +637,53 @@ public override void Dispose()
...
-        private string ResolvePluginPath(string path) =>
-            Uri.IsWellFormedUriString(path, UriKind.Absolute)
-                ? DownloadPlugin(path).Result
-                : path;
+        private string ResolvePluginPath(string path)
+        {
+            if (Uri.IsWellFormedUriString(path, UriKind.Absolute))
+            {
+                // Run the download on a background thread
+                return Task.Run(() => DownloadPlugin(path, SwarmCancellationToken)).GetAwaiter().GetResult();
```
moreal, line 645: Using `SwarmCancellationToken` here seems invalid. Because `SwarmCancellationToken` is Initialized in `ExecuteAsync -> StartSwarm`. But `ResolvePluginPath` is called in `LibplanetNodeService`'s constructor. 

This may be a sign that you need to separate the downloading logic from LibplanetNodeService.

Libplanet.Headless/Hosting/LibplanetNodeService.cs
```diff
@@ -635,29 +637,53 @@ public override void Dispose()
...
+                var content = await httpClient.GetByteArrayAsync(url, cancellationToken);
+                await File.WriteAllBytesAsync(downloadPath, content, cancellationToken);
+                logger.Debug("Finished downloading.");
+            }
+            else
+            {
+                logger.Debug("Download skipped, file already exists.");
+            }
+
+            if (!Directory.Exists(extractPath))
```
moreal, line 676: In the future, it needs to check checksum of directory's files. This is because if you unzip during a previous run and get hung up in the middle, you'll end up with an incomplete directory, and you can't check for this error just by looking at the existence of the directory.

area363: Could you check https://github.com/planetarium/NineChronicles.Headless/pull/2456/commits/2e06152c012211aa8a60d44da0d5169b1f42e34c if this would work? 🙏

Libplanet.Headless/Hosting/LibplanetNodeService.cs
area363: Hmm separating the downloading logic from LibplanetNodeService seems a bit tricky since the ResolvePluginPath(https://github.com/planetarium/NineChronicles.Headless/blob/development/Libplanet.Headless/Hosting/LibplanetNodeService.cs#L126) is called within the LibplanetNodeService constructor. 🤔 

I've removed the wrong use of SwarmCancellationToken entirely for now but do you have any suggestions?

moreal: For now, I don't have anything else to suggest in this PR: this repository is in need of an overall refactoring. LibplanetNodeService is also code that I wrote when I didn't know .NET DI very well and I think it's actually better to have a generate DI logic that creates an IActionEvaluator and use that IActionEvaluator when generating BlockChain.

Libplanet.Headless/Hosting/LibplanetNodeService.cs
moreal: About this:

- If `checksum.txt` is in `extractPath`, the checksum which made of the files in `extractPath`, will be different. Because, when the first running, `checksum.txt = md5(filesInZip)` but when the second running, `checksum.txt = md5(filesInZip + checksum.txt)`. The checksum must be calculated from the provider side (e.g., GitHub Actions).
- Separately, we talked about checksums, but I think it's okay to simply use the completion flag after decompression. Of course, you still need a checksum to check if the DLL is corrupted or something, and as mentioned above, the provider should provide it.

Libplanet.Headless/Hosting/LibplanetNodeService.cs
area363: Ah okok, I understand 🙆‍♂️

Libplanet.Headless/Hosting/LibplanetNodeService.cs
area363: Let me know if https://github.com/planetarium/NineChronicles.Headless/commit/2e06152c012211aa8a60d44da0d5169b1f42e34c seems ok to you and I'll finalize this PR!

PR #2477, opened 2024-05-16
Provide `index` parameter when to query state

moreal:
## Description

When it's a PoW chain, it's very risky to access state by index. However, thanks to the work of the Libplanet team, the NineChronicles network has partially transitioned to PoS, so we're adding the ability to access it by index. We expect that this will solve the problem of replacing index with hash and then accessing it.

## GraphQL

```graphql
{
  state(address: "0x", index: 0)
  stateQuery(index: 0)
  {
    avatar(address: "0x") { name }
  }
}
```


boscohyun (approved):
It's good to apply it to unit tests as well.

U-lis (approved):
LGTM

PR #2479, opened 2024-05-21
Implement diff query

Atralupus:
Adding State Diff Query for Storing Mimir Data

### DIFF
--- a/NineChronicles.Headless/GraphTypes/Diff/RootStateDiffType.cs
+++ b/NineChronicles.Headless/GraphTypes/Diff/RootStateDiffType.cs
@@ -25,7 +25,7 @@ public class RootStateDiffType : ObjectGraphType<RootStateDiffType.Value>
             description: "The path to the root state difference."
         );
 
-        Field<ListGraphType<StateDiffType>>(
+        Field<NonNullGraphType<ListGraphType<NonNullGraphType<StateDiffType>>>>(
             "Diffs",
             description: "List of state differences under this root."
         );
--- a/NineChronicles.Headless/GraphTypes/Diff/RootStateDiffType.cs
+++ b/NineChronicles.Headless/GraphTypes/Diff/RootStateDiffType.cs
@@ -1,6 +1,6 @@
 using GraphQL.Types;
 
-namespace NineChronicles.Headless.GraphTypes.States;
+namespace NineChronicles.Headless.GraphTypes.Diff;
 
 public class RootStateDiffType : ObjectGraphType<RootStateDiffType.Value>
 {
--- a/NineChronicles.Headless/GraphTypes/Diff/RootStateDiffType.cs
+++ b/NineChronicles.Headless/GraphTypes/Diff/RootStateDiffType.cs
@@ -4,7 +4,7 @@ namespace NineChronicles.Headless.GraphTypes.Diff;
 
 public class RootStateDiffType : ObjectGraphType<RootStateDiffType.Value>
 {
-    public class Value
+    public class Value : IDiffType
     {
         public string Path { get; }
         public StateDiffType.Value[] Diffs { get; }
--- a/NineChronicles.Headless/GraphTypes/Diff/StateDiffType.cs
+++ b/NineChronicles.Headless/GraphTypes/Diff/StateDiffType.cs
@@ -10,10 +10,10 @@ public class StateDiffType : ObjectGraphType<StateDiffType.Value>
     public class Value
     {
         public string Path { get; }
-        public IValue? BaseState { get; }
+        public IValue BaseState { get; }
         public IValue? ChangedState { get; }
 
-        public Value(string path, IValue? baseState, IValue? changedState)
+        public Value(string path, IValue baseState, IValue? changedState)
         {
             Path = path;
             BaseState = baseState;
@@ -32,13 +32,11 @@ public class StateDiffType : ObjectGraphType<StateDiffType.Value>
             description: "The path of the state difference."
         );
 
-        Field<StringGraphType>(
+        Field<NonNullGraphType<StringGraphType>>(
             "BaseState",
             description: "The base state before changes.",
             resolve: context =>
-                context.Source.BaseState is null
-                    ? null
-                    : ByteUtil.Hex(new Codec().Encode(context.Source.BaseState))
+                ByteUtil.Hex(new Codec().Encode(context.Source.BaseState))
         );
 
         Field<StringGraphType>(
--- a/NineChronicles.Headless/GraphTypes/Diff/StateDiffType.cs
+++ b/NineChronicles.Headless/GraphTypes/Diff/StateDiffType.cs
@@ -3,7 +3,7 @@ using Bencodex.Types;
 using GraphQL.Types;
 using Libplanet.Common;
 
-namespace NineChronicles.Headless.GraphTypes.States;
+namespace NineChronicles.Headless.GraphTypes.Diff;
 
 public class StateDiffType : ObjectGraphType<StateDiffType.Value>
 {
@@ -19,8 +19,6 @@ public class StateDiffType : ObjectGraphType<StateDiffType.Value>
             BaseState = baseState;
             ChangedState = changedState;
         }
-
-        public bool IsAddressRelated(string address) => Path.Contains(address);
     }
 
     public StateDiffType()
--- a/NineChronicles.Headless/GraphTypes/Diff/StateDiffType.cs
+++ b/NineChronicles.Headless/GraphTypes/Diff/StateDiffType.cs
@@ -7,7 +7,7 @@ namespace NineChronicles.Headless.GraphTypes.Diff;
 
 public class StateDiffType : ObjectGraphType<StateDiffType.Value>
 {
-    public class Value
+    public class Value : IDiffType
     {
         public string Path { get; }
         public IValue BaseState { get; }
@@ -33,8 +33,7 @@ public class StateDiffType : ObjectGraphType<StateDiffType.Value>
         Field<NonNullGraphType<StringGraphType>>(
             "BaseState",
             description: "The base state before changes.",
-            resolve: context =>
-                ByteUtil.Hex(new Codec().Encode(context.Source.BaseState))
+            resolve: context => ByteUtil.Hex(new Codec().Encode(context.Source.BaseState))
         );
 
         Field<StringGraphType>(
--- a/NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
+++ b/NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
@@ -77,12 +77,12 @@ namespace NineChronicles.Headless.GraphTypes
             Field<ListGraphType<RootStateDiffType>>(
                 name: "diffs",
                 arguments: new QueryArguments(
-                    new QueryArgument<LongGraphType>
+                    new QueryArgument<NonNullGraphType<LongGraphType>>
                     {
                         Name = "baseIndex",
                         Description = "The index of the block used to fetch state from chain."
                     },
-                    new QueryArgument<LongGraphType>
+                    new QueryArgument<NonNullGraphType<LongGraphType>>
                     {
                         Name = "changedIndex",
                         Description = "The index of the block used to fetch state from chain."
@@ -97,24 +97,20 @@ namespace NineChronicles.Headless.GraphTypes
                         );
                     }
 
-                    var baseIndex = context.GetArgument<long?>("baseIndex");
-                    var changedIndex = context.GetArgument<long?>("changedIndex");
+                    var baseIndex = context.GetArgument<long>("baseIndex");
+                    var changedIndex = context.GetArgument<long>("changedIndex");
 
-                    long latestBlockIndex = blockChain.Tip.Index;
-                    var filledChangedIndex = baseIndex ?? latestBlockIndex;
-                    var filledBaseIndex = changedIndex ?? filledChangedIndex - 1;
+                    var blockInterval = changedIndex - baseIndex;
 
-                    var baseBlockStateRootHash = blockChain[filledBaseIndex].StateRootHash.ToString();
-                    var changedBlockStateRootHash = blockChain[filledChangedIndex].StateRootHash.ToString();
-
-                    var blockInterval = filledChangedIndex - filledBaseIndex;
-
-                    if (blockInterval >= 10)
+                    if (blockInterval >= 10 && blockInterval > 0)
                     {
                         throw new ExecutionError(
-                            "Block interval should be under 10"
+                            "Interval between baseIndex and changedIndex should under 10"
                         );
                     }
+
+                    var baseBlockStateRootHash = blockChain[baseIndex].StateRootHash.ToString();
+                    var changedBlockStateRootHash = blockChain[changedIndex].StateRootHash.ToString();
 
                     var baseStateRootHash = HashDigest<SHA256>.FromString(baseBlockStateRootHash);
                     var targetStateRootHash = HashDigest<SHA256>.FromString(
--- a/NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
+++ b/NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
@@ -25,6 +25,8 @@ using Nekoyume.Module;
 using NineChronicles.Headless.GraphTypes.States;
 using static NineChronicles.Headless.NCActionUtils;
 using Transaction = Libplanet.Types.Tx.Transaction;
+using System.Security.Cryptography;
+using System.Text;
 
 namespace NineChronicles.Headless.GraphTypes
 {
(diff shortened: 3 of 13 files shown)


NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
```diff
@@ -72,6 +74,91 @@ public StandaloneQuery(StandaloneContext standaloneContext, IConfiguration confi
...
+
+                    var baseBlockStateRootHash = blockChain[filledBaseIndex].StateRootHash.ToString();
+                    var changedBlockStateRootHash = blockChain[filledChangedIndex].StateRootHash.ToString();
+
+                    var blockInterval = filledChangedIndex - filledBaseIndex;
+
+                    if (blockInterval >= 10)
+                    {
+                        throw new ExecutionError(
+                            "Block interval should be under 10"
```
moreal, line 115:
```suggestion
                            "Interval between baseIndex and changedIndex should under 10"
```

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
```diff
@@ -72,6 +74,91 @@ public StandaloneQuery(StandaloneContext standaloneContext, IConfiguration confi
                 }
             );
 
+            Field<ListGraphType<RootStateDiffType>>(
+                name: "diffs",
```
moreal, line 78: I personally don't like it to work implicitly, I think it's better to accept them all as Required instead of Nullable, but it's okay if they're not accepted. If you want to give a default, please write about the default behavior in the GraphQL description.

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
```diff
@@ -72,6 +74,91 @@ public StandaloneQuery(StandaloneContext standaloneContext, IConfiguration confi
                 }
             );
 
+            Field<ListGraphType<RootStateDiffType>>(
```
moreal, line 77:
```suggestion
            Field<NonNullGraphType<ListGraphType<NonNullGraphType<RootStateDiffType>>>>(
```

NineChronicles.Headless/GraphTypes/States/StateDiffType.cs
```diff
@@ -0,0 +1,53 @@
...
+using Libplanet.Common;
+
+namespace NineChronicles.Headless.GraphTypes.States;
+
+public class StateDiffType : ObjectGraphType<StateDiffType.Value>
+{
+    public class Value
+    {
+        public string Path { get; }
+        public IValue? BaseState { get; }
```
moreal, line 13: Maybe `BaseState` always be non-nullable.

```suggestion
        public IValue? BaseState { get; }
```

moreal (commented):
Since `RootStateDiffType.Value` and `StateDiffType.Value` doesn't mean in-game state, it may be better to place other directory 🤔

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
Atralupus: 👍

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
Atralupus: I did my best to match the setup of EmptyChronicles.
I got rid of nulls in responses and made all arguments mandatory. Please check it

NineChronicles.Headless/GraphTypes/Diff/RootStateDiffType.cs
```diff
@@ -0,0 +1,33 @@
+using GraphQL.Types;
+
+namespace NineChronicles.Headless.GraphTypes.States;
```
moreal, line 3:
```suggestion
namespace NineChronicles.Headless.GraphTypes.Diff;
```

NineChronicles.Headless/GraphTypes/Diff/StateDiffType.cs
```diff
@@ -0,0 +1,51 @@
+using Bencodex;
+using Bencodex.Types;
+using GraphQL.Types;
+using Libplanet.Common;
+
+namespace NineChronicles.Headless.GraphTypes.States;
```
moreal, line 6:
```suggestion
namespace NineChronicles.Headless.GraphTypes.Diff;
```

NineChronicles.Headless/GraphTypes/Diff/StateDiffType.cs
```diff
@@ -0,0 +1,51 @@
...
+        public IValue? ChangedState { get; }
+
+        public Value(string path, IValue baseState, IValue? changedState)
+        {
+            Path = path;
+            BaseState = baseState;
+            ChangedState = changedState;
+        }
+
+        public bool IsAddressRelated(string address) => Path.Contains(address);
```
moreal, line 23: It seems unnecessary.

```suggestion
        }
```

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
```diff
@@ -72,6 +74,87 @@ public StandaloneQuery(StandaloneContext standaloneContext, IConfiguration confi
...
+                            $"{nameof(StandaloneContext)}.{nameof(StandaloneContext.BlockChain)} was not set yet!"
+                        );
+                    }
+
+                    var baseIndex = context.GetArgument<long>("baseIndex");
+                    var changedIndex = context.GetArgument<long>("changedIndex");
+
+                    var blockInterval = changedIndex - baseIndex;
+
+                    if (blockInterval >= 10 && blockInterval > 0)
```
moreal: Maybe it means:

```suggestion
                    var blockInterval = Math.abs(changedIndex - baseIndex);

                    if (blockInterval >= 10)
```

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
```diff
@@ -72,6 +74,87 @@ public StandaloneQuery(StandaloneContext standaloneContext, IConfiguration confi
...
+                    {
+                        throw new ExecutionError(
+                            "Interval between baseIndex and changedIndex should under 10"
+                        );
+                    }
```
boscohyun, line 110: Did you intend like below?
```suggestion
                    var blockInterval = changedIndex - baseIndex;
                    if (blockInterval < 0 || blockInterval >= 10)
                    {
                        throw new ExecutionError(
                            "Interval between baseIndex and changedIndex should be greater than or equal to 0 and less than 10."
                        );
                    }
```

Atralupus: Thanks, done!

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
```diff
@@ -72,6 +74,87 @@ public StandaloneQuery(StandaloneContext standaloneContext, IConfiguration confi
...
+                    new QueryArgument<NonNullGraphType<LongGraphType>>
+                    {
+                        Name = "baseIndex",
+                        Description = "The index of the block used to fetch state from chain."
+                    },
+                    new QueryArgument<NonNullGraphType<LongGraphType>>
+                    {
+                        Name = "changedIndex",
+                        Description = "The index of the block used to fetch state from chain."
+                    }
```
boscohyun, line 89: I've added some description to the arguments.🙂
```suggestion
                arguments: new QueryArguments(
                    new QueryArgument<NonNullGraphType<LongGraphType>>
                    {
                        Name = "baseIndex",
                        Description = "The index of the base block used to fetch state from chain." +
                                      "It should be less than or equal to the chain tip index" +
                                      " and greater than or equal to the genesis block index(0)" +
                                      " and less than the changedIndex."
                    },
                    new QueryArgument<NonNullGraphType<LongGraphType>>
                    {
                        Name = "changedIndex",
                        Description = "The index of the target block used to fetch state from chain." +
                                      "It should be less than or equal to the chain tip index" +
                                      " and greater than or equal to the genesis block index(0)." +
                                      " and greater than the baseIndex."
                    }
```

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
```diff
@@ -72,6 +74,87 @@ public StandaloneQuery(StandaloneContext standaloneContext, IConfiguration confi
...
+                    var targetTrieModel = stateStore.GetStateRoot(targetStateRootHash);
+
+                    var diffs = baseTrieModel
+                        .Diff(targetTrieModel)
+                        .Select(x =>
+                        {
+                            var baseSubTrieModel = stateStore.GetStateRoot(
+                                new HashDigest<SHA256>((Binary)x.SourceValue)
+                            );
+                            if (x.TargetValue is not null)
```
moreal, line 131: We should handle the case when `TargetValue` is `null`, too. Maybe we should resolve issue with the *union* type.

In TypeScript:

```
// If TargetValue is not null, it can be root of subtrie. => { path: string, diffs: StateDiffType[] }
// If TargetValue is null, it means the subtrie has removed. => StateDiffType.
type RootStateDiffType = { path: string; diffs: StateDiffType[] } | StateDiffType;
type StateDiffType = { path: string; baseState: unknown; changedState: unknown };
```

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
boscohyun: Why not increase the minimum value of `blockInterval` from `0` to `1`?
I thought if `blockInterval` was `0`, there would be no diff.

NineChronicles.Headless/GraphTypes/Diff/RootStateDiffType.cs
boscohyun: `�`🙃

Atralupus: @.@ thanks, my mistake

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
moreal: Since there are no limits for `changedIndex`, `baseIndex`, the `blockInterval` can be negative. I think it is normal case.

NineChronicles.Headless/GraphTypes/Diff/StateDiffType.cs
Atralupus: yes right 👍 thanks

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
boscohyun: Oh, and it also provides a lookup for reverse changes, right?

moreal: Yes, you're right. It is just trie diff.

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
Atralupus: I figured that if `baseIndex` is higher than `changedIndex`, it’s a mistake.
So, I made sure the validation doesn’t allow negatives.

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
boscohyun: If so, I'll have to write it differently, not as suggested in this thread.
I'll also need to revise the description of the query arguments.🙂

Atralupus: Even if 0 is allowed, the GraphQL server ultimately does nothing and just returns an empty list.
Therefore, I thought it should not be allowed in the validation phase.

boscohyun: > Even if 0 is allowed, the GraphQL server ultimately does nothing and just returns an empty list.
Therefore, I thought it should not be allowed in the validation phase.

If so, my suggestion above would be valid.

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
Atralupus: https://github.com/planetarium/NineChronicles.Headless/pull/2479#discussion_r1609426393
My bad, I thought negative values were not acceptable.

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
Atralupus: > Why not increase the minimum value of `blockInterval` from `0` to `1`? I thought if `blockInterval` was `0`, there would be no diff.

Sorry I misunderstood, However changed the rule, please check it

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
Atralupus: Applied, thanks

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
boscohyun: > Sorry I misunderstood, However changed the rule, please check it

Oh, never mind.😉

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
```diff
@@ -72,6 +75,94 @@ public StandaloneQuery(StandaloneContext standaloneContext, IConfiguration confi
...
+                    new QueryArgument<NonNullGraphType<LongGraphType>>
+                    {
+                        Name = "changedIndex",
+                        Description = "The index of the target block used to fetch state from the chain. " +
+                                    "It should be less than or equal to the chain tip index, " +
+                                    "greater than or equal to the genesis block index (0), " +
+                                    "and greater than baseIndex. The difference between baseIndex " +
+                                    "and changedIndex should not be greater than 10 and should not be zero."
+                    }
+                ),
```
boscohyun, line 99: The argument descriptions need to fix.
Could you add description for "diffs" query?

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
Atralupus: It's meant to be like this, right?

```
var diffs = baseTrieModel
    .Diff(targetTrieModel)
// Type error
    .Select(x =>
    {
        if (x.TargetValue is not null)
        {
            var targetSubTrieModel = stateStore.GetStateRoot(new HashDigest<SHA256>((Binary)x.TargetValue));
            return new RootStateDiffType.Value(
                Encoding.Default.GetString(x.Path.ByteArray.ToArray()),
                targetSubTrieModel.Diff(baseTrieModel)
                    .Select(diff => new StateDiffType.Value(
                        Encoding.Default.GetString(diff.Path.ByteArray.ToArray()),
                        diff.SourceValue,
                        diff.TargetValue
                    )).ToArray()
            );
        }
        else
        {
            return new StateDiffType.Value(
                Encoding.Default.GetString(x.Path.ByteArray.ToArray()),
                x.SourceValue,
                x.TargetValue
            );
        }
    }).ToArray();
```

Atralupus: From what I understand, it's tough to use union types directly in .NET (thankfully, GraphQL has them).
So, for the 'diffs' variable, it seems like we need to use a sort of fake, upper-level interface and handle it that way. Does that sound right?

```csharp
var diffs = baseTrieModel
    .Diff(targetTrieModel)
    .Select(x =>
    {
        if (x.TargetValue is not null)
        {
            var targetSubTrieModel = stateStore.GetStateRoot(new HashDigest<SHA256>((Binary)x.TargetValue));
            return (IDiffResult)new RootStateDiffTypeValue(
                Encoding.Default.GetString(x.Path.ByteArray.ToArray()),
                targetSubTrieModel.Diff(baseTrieModel)
                    .Select(diff => new StateDiffTypeValue(
                        Encoding.Default.GetString(diff.Path.ByteArray.ToArray()),
                        diff.SourceValue,
                        diff.TargetValue
                    )).ToArray()
            );
        }
        else
        {
            return (IDiffResult)new StateDiffTypeValue(
                Encoding.Default.GetString(x.Path.ByteArray.ToArray()),
                x.SourceValue,
                x.TargetValue
            );
        }
    }).ToArray();
```

moreal: Yes, you can see https://graphql-dotnet.github.io/docs/getting-started/unions/

Atralupus: Done!!!!!!!

NineChronicles.Headless/GraphTypes/StandaloneQuery.cs
Atralupus: yes roger that

moreal (approved):
Awesome GraphQL field document!! 👍🏻 👍🏻 👍🏻  @boscohyun @Atralupus

boscohyun (approved):

PR #2494, opened 2024-07-02
Bump Lib9c to 1.14.0

OnedgeLee:
### Lib9c Changes

- Fix not to use GetNextWorldState() on tx validation
- Reduce cache size of HashNodeCache


OnedgeLee:
Duplicated PR, Closing

PR #2509, opened 2024-07-24
Add txStatusFilter in ncTransactions type

Atralupus:
Currently, checking the transaction status using the transaction result query consumes unnecessary network resources.
To address this, I propose adding a txStatus filter argument to the ncTransactions type.

### DIFF
--- a/NineChronicles.Headless/GraphTypes/TransactionHeadlessQuery.cs
+++ b/NineChronicles.Headless/GraphTypes/TransactionHeadlessQuery.cs
@@ -77,7 +77,9 @@ namespace NineChronicles.Headless.GraphTypes
                     new QueryArgument<NonNullGraphType<LongGraphType>>
                     { Name = "limit", Description = "number of block to query." },
                     new QueryArgument<NonNullGraphType<StringGraphType>>
-                    { Name = "actionType", Description = "filter tx by having actions' type. It is regular expression." }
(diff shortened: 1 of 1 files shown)


NineChronicles.Headless/GraphTypes/TransactionHeadlessQuery.cs
```diff
@@ -101,6 +104,15 @@ public TransactionHeadlessQuery(StandaloneContext standaloneContext)
                                 return false;
                             }
 
+                            if (txStatusFilter is not null)
+                            {
+                                var txResult = TxResult(standaloneContext, tx.Id) as TxResult;
+                                if (txResult is not null && !txStatusFilter.Contains(txResult.TxStatus))
```
moreal, line 110: We should think about the situation where there is no `TxResult`. If `txStatusFilter` is given but there is no `txResult` for txs, shouldn't we ignore the transaction? Or not?

moreal (commented):
Just an opinion.

NineChronicles.Headless/GraphTypes/TransactionHeadlessQuery.cs
Atralupus: @.@ We can ignore this issue since txStatusFilter is nullable, which was designed to handle cases where there is no txResult.

boscohyun (approved):

moreal (approved):

PR #2516, opened 2024-07-30
Bump OpenTelemetry dependencies

moreal:
Currently, Prometheus fails to fetch metrics from headless because its metrics export result isn't compatible for the standard or Prometheus' way. So I had disabled it in #2460 temporarily. And it is enabled again in this version and the error occurs again. But, fortunately, it seems fixed in OpenTelemetry 1.9.0 by https://github.com/open-telemetry/opentelemetry-dotnet/pull/5646. So this pull request bumps them to 1.9.0

```
unit not a suffix of metric "process_runtime_dotnet_gc_allocations_size_bytes_total"
```


area363 (approved):

s2quake:
@area363 Thanks for merging it on my behalf.

PR #2544, opened 2024-08-22
Fix/tracing

sky1045:
Add trace for graphql query
Fix ActivitySource as a static member


moreal (approved):

riemannulus:
Not only about `ActivitySource`, but also an adjustment to `Meter`. Because `ActivitySource` and `Meter` are `IDisposable`, they should be controlled by following practices.
- Make it static and do not release or dispose of it until the process is stopped.
- Call the `dispose` by each using `Metric` and `Trace`.
- Using DI for `Meter` and `ActivitySource` and dispose of it by ASP.NET 

`Meter` is not a context in this PR, but hopefully, we should fix `Meter` too.

PR #2563, opened 2024-08-30
Introduce `filtered` argument at `.nodeStatus.stagedTxIds[Count]`

moreal:
Since `stagedTxIds` and `stagedTxIdsCount` field uses `BlockChain.GetStagedTransactionIds` method, it uses `IStagePolicy.Iterate(BlockChain chain, bool filtered)` with `filtered=true`. But we cannot know how many real transactions are staged now because it must not filter transactions to know that.

This pull request replaces `BlockChain.GetStagedTransactionIds()` call with `IStagePolicy.Iterate(...)` and introduces `filtered` query argument for `.nodeStatus.stagedTxIds` and `.nodeStatus.stagedTxIdsCount` fields.

### DIFF
--- a/NineChronicles.Headless/GraphTypes/NodeStatus.cs
+++ b/NineChronicles.Headless/GraphTypes/NodeStatus.cs
@@ -87,6 +87,12 @@ namespace NineChronicles.Headless.GraphTypes
                     {
                         Name = "address",
                         Description = "Target address to query"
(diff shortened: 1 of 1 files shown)


ipdae:
이건 @libplanet 팀에서 봐주시는게 좋을것같습니다.

PR #2583, opened 2024-09-13
Fix broken tests

U-lis:
AvatarState does not receive AvatarSheets anymore.


moreal (approved):
You can use just `AvatarState.Create` static method, same method signature with the previous `AvatarState` constructor. (Because there are already sheets)

PR #2611, opened 2024-10-27
chore: clean arenaParticipantsSync option

### LINKED ISSUE #2598: Deprecate `--arenaParticipantsSync` option
## Context

Since #2596, `--arenaParticipantsSync` CLI option becomes not to be used. So it should be deprecated.

## Work items

Please notice users to know that "the option is deprecated so you shouldn't use the option anymore", through stderr output.

## Tips

Since `arenaParticipantsSync` default value is `true`, you should check the case where `arenaParticipantsSync` is `false`.

Gongkirby:
Resolve #2598


riemannulus (approved):
LGTM!

PR #2620, opened 2024-11-11
chore: remove pre-restoring from Dockerfiles

moreal:


U-lis (commented):
How to restore all the packages if we delete all the copy and restore commands?
Does dotnet do it automatically?

moreal:
> How to restore all the packages if we delete all the copy and restore commands? Does dotnet do it automatically?

@U-lis .NET restores dependencies automatically when building project.

U-lis (approved):

PR #2625, opened 2024-11-11
chore: update dependabot conf to bump GitHub Actions automatically

moreal:
We usually missed to bump GitHub Actions and cannot realizing if there is a new release to bump. Since Dependabot supports GitHub Actions, I updated its configuration to check if there is something to bump automatically 🤖

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,8 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
+      interval: "weekly"


U-lis (commented):
Is it necessary to check on github actions?
I think this is good-to-be, not must-to-be.

moreal:
> Is it necessary to check on github actions? I think this is good-to-be, not must-to-be.

@U-lis I'm not sure if I understand correctly, but Dependabot doesn't run on GitHub Actions - it's a separate behavior from events like opening a PR, doing a push, etc.

Or do I really need a workflow that checks if the GitHub Actions workflow syntax is correct?

U-lis:
Ah, just my bad. I misunderstood this PR at first glance.
Looks good to me :+1:

U-lis (approved):

